PR #51, opened 2022-11-22
Better symmetry

fkraushofer:
Fundamentally changes the algorithm used to detect plane symmetry. The old algorithm scales with O(n^2), this new version scales with O(n), where n is the number of atoms in the lowest-occupation sublayer.

In the old version, candidate points for rotations/mirrors/glides were generated from pairs and trios of atoms on a (2x2) supercell. Then, all candidate symmetry operations were tested in these positions. In the new algorithm, each symmetry operation is applied to the slab just once, then the slab is tested for 2D translation symmetry. The only candidates for translation symmetry are atom-to-atom translations, and it is sufficient to test translation vectors linking an arbitrary atom in the lowest-occupancy sublayer to every atom in that sublayer (including itself). The position of the rotation axis or mirror/glide plane can then be determined based on the translation vector.

Also fixed several small bugs with bulk symmetry detection:
- bulk doubling was not always performed when required, resulting in incorrect bulk symmetry identifications when the symmetry search ran on a slab consisting of only one layer
- bulk translation symmetry was compared "element-blind", so e.g. ordered alloys were assigned higher symmetry than they actually have.

### DIFF
--- a/tests/test_beamgen.py
+++ b/tests/test_beamgen.py
@@ -63,7 +63,7 @@ class TestGenerateBeamlist:
 
     @fixture(name='make_beamlist')
     @parametrize_with_cases('args', **_BEAMGEN_CASES)
-    def fixture_make_beamlist(self, args, tmp_path_factory):
+    def fixture_make_beamlist(self, args, tmp_path_factory, tensorleed_path):
         """Return slab, parameters, info and the path to a 'BEAMLIST'."""
         slab, param, info = args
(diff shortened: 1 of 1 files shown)


fkraushofer:
General question: What's the status of this, and is it theoretically possible to merge it into the dev/try-me branch? It could probably profit from some real-world testing before being merged into master. I remember it was waiting for some of the class refactors, but have no idea whether this is still true.

PR #148, opened 2023-12-19
Advanced Full Dynamic Optimization

### LINKED ISSUE #96: Viperleed, Error plots of a and b using Full-dynamic Opitmization
When performing a full dynamic optimisation to obtain values for the in plane lattice vectors, it would be interesting to know how defective the fitted values are. Therefor, it would be useful to to create the same errorplot.pdf as out of Run 5, with R + varR as line in the parabolic fit.

alexmiame:
This PR is a project of mine I've been working on for a bit. It implements a rework and extension of the full-dynamic calculation and enables a more generalized optimization:

- The full dynamic optimization is refactored to internally use generalized and modular optimizers.
- This enables concurrent optimization of multiple parameters (e.g. incidence angles theta and phi together).
- Single and multi parameter optimizations can use any of the non-gradient based algorithms implemented by scipy.optimize.
- Single parameters can use the parabola fit optimization as before.
- Single parameters can now be scanned in a brute force fashion with a user-defined range and step-width. This produces plots with error estimates like in the error-calculation, but extended to full-dynamic parameters. (Closes #96)
- All optimizers output a CSV file of the sampled parameter values and R-factors for custom post-processing by users.

It would also be fairly easy to extend this new optimization to access structural parameters, i.e. displacements, vibrations and occupations. This would enable full dynamic structural optimizations, which may be interesting in the case of very small unit cells or for a final very-fine optimization of a converged structure.

The PR still needs some work, but most parts are already functional, though maybe in a rough state.
In particular, there are a still a few TODOs to take care of, plus it needs tests and updates to the documentation.


amimre:
I'm converting this PR to a draft for now, as I'm not sure which version it will go into.
Also, we probably will want to have some discussion about harmonizing these FD optimizers with the ones for the new tensor-LEED implementation.

PR #210, opened 2024-06-26
Rename PatternInfo.tlm to experiment_symmetry.ini

amimre:


michele-riva:
Thanks for adding the tests! I would still hold off with this one, though, as we need to support the .ini format in the GUI. That's my next task after #198.

fkraushofer (approved):
Not actually reviewed, approving blind from my phone to circumvent the restriction - @amimre please make the decision whether it's ready to merge or not.

amimre:
I found one missing backtick in the docs page and fixed it. Good to go otherwise from my side.
@michele-riva you should be able to review and merge it now.

PR #230, opened 2024-07-09
Building docs on Readthedocs.org

amimre:
This PR adds the `.yaml` file needed to build on buildthedocs.org. A preliminary version can be accessed on [https://viperleed.readthedocs.io/en/latest/index.html](https://viperleed.readthedocs.io/latest/index.html) already. Once merged, we can transfer the [viperleed.org](url) domain to readthedocs.


michele-riva:
Cool, thanks! I was wondering if we even actually need to use readthedocs, and whether we could not just use viperleed.org for hosting. I did not really research much, but will using readthedocs create a duplicate documentation?

michele-riva:
Also, I'm looking into [nox](https://github.com/wntrblm/nox) right now, and it looks like we could use it for everything: linting, testing, making docs. I just started with linting for now, as it seemed a bit easier (branch yet to be published).

michele-riva:
Alright, if there is no duplication then I'm fine with it. Yes, let's go for the ad-less version. Could you get a quote/invoice to Stefan?

michele-riva:
Sorry, just to make sure: will we still have access to the SSL support and the extra emails on viperleed.org? Or are you suggesting to keep both plans?

amimre:
> Sorry, just to make sure: will we still have access to the SSL support and the extra emails on viperleed.org? Or are you suggesting to keep both plans?

The SSL is not necessary since it's not going to be hosted by the old hosting service. readthedocs includes SSL by default.
If we want the extra email addresses, we can get an email only package from our current (or any other) hosting provider, which is much cheaper than the full web hosting package.

michele-riva:
Sounds like a plan. I'd have the email package from wherever it is most convenient, as long as they're @viperleed.org

.readthedocs.yaml
```diff
@@ -0,0 +1,28 @@
...
+  tools:
+    python: "3.12"
+
+# Build documentation in the "doc/" directory with Sphinx
+sphinx:
+  configuration: doc/conf.py
+
+# Optionally build your docs in additional formats such as PDF and ePub
+# formats:
+#    - pdf
```
michele-riva, line 20: I think we can do also PDF. Readthedocs will display a download button, if I got it right.

.readthedocs.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# install the dependencies
+python:
+   install:
```
michele-riva, line 24: Don't we need the checkout action somewhere here?

.readthedocs.yaml
```diff
@@ -0,0 +1,28 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
```
michele-riva, line 5: I think you could make this "dependent" on the [distribute_built_dist.yml](https://github.com/viperleed/viperleed/blob/github-workflow-dev/.github/workflows/distribute_built_dist.yml) workflow that we will run automatically (after testing) every time we make a new release. I don't think it makes much sense to have documentation online for yet-to-be-released stuff.

michele-riva (changes requested):
Wonderful. A few suggestions.

.readthedocs.yaml
amimre: Yes, the issue is the pdf is too large to automatically build right now. Maybe the quota is increased with the paid version automatically, otherwise we can ask readthedocs to do it (see comment in [https://github.com/readthedocs/readthedocs.org/issues/8382](https://github.com/readthedocs/readthedocs.org/issues/8382)).

.readthedocs.yaml
amimre: Sure, we can select the branch from which to build in the admin panel.

.readthedocs.yaml
michele-riva: I see

.readthedocs.yaml
michele-riva: Right now the action I have created builds only from master, which I'd always keep to date with the most recent version, but perhaps you're right, that we will eventually need other branches too

.readthedocs.yaml
michele-riva: Now that I had a closer look at the issue, I have had the problem myself early on when trying to build the documentation. I think that's the reason why I have been building the pdf on Windows. MikTeX does not seem to have the problem. I think the problem in the issue was the builder itself. We also switched to xelatex, probably partly because of the same problem.

amimre:
I got around to looking at this again. I've brought this PR back in sync with the current state of the repo and fixed some things.
The PDF build is still broken in a way that I cannot reproduce locally. I contacted the readthedocs support.

michele-riva:
> I got around to looking at this again. I've brought this PR back in sync with the current state of the repo and fixed some things. The PDF build is still broken in a way that I cannot reproduce locally. I contacted the readthedocs support.

Thanks for updating this. I saw that an email came from them earlier today. Let me know if you didn't get it.

amimre:
I managed to fix the issue now. It turned out to be a missing dependency installation (Inkscape) for the conversion of SVG to PDF for the figures we use.
The preview of the readthedocs build is up here, now including a download button for the PDF version: [https://viperleed.readthedocs.io/en/latest/](https://viperleed.readthedocs.io/en/latest/)
We are only showing the latest version now, but we could easily make older versions of the documentation available by back-porting the `.readthedocs.yaml` file.

With this fixed, I'd argue that we make the move to the hosting on readthedocs as soon as possible, since the current online docs are already a bit outdated.

michele-riva:
> I managed to fix the issue now. It turned out to be a missing dependency installation (Inkscape) for the conversion of SVG to PDF for the figures we use.

Wonderful, thanks for fixing this!

> The preview of the readthedocs build is up here, now including a download button for the PDF version: https://viperleed.readthedocs.io/latest/ We are only showing the latest version now, but we could easily make older versions of the documentation available by back-porting the `.readthedocs.yaml` file.

Good to know. Perhaps worth looking into this.

> With this fixed, I'd argue that we make the move to the hosting on readthedocs as soon as possible, since the current online docs are already a bit outdated.

I just noticed that the readthedocs-hosted documentation contains ads. Is there a way to prevent this? I'm not a fan of this, and our current hosting package includes no ads. We can easily switch from paying our current host to paying readthedocs.

michele-riva:
> > We are only showing the latest version now, but we could easily make older versions of the documentation available by back-porting the `.readthedocs.yaml` file.
> 
> Good to know. Perhaps worth looking into this.
> 
> > With this fixed, I'd argue that we make the move to the hosting on readthedocs as soon as possible, since the current online docs are already a bit outdated.
> 
> I just noticed that the readthedocs-hosted documentation contains ads. Is there a way to prevent this? I'm not a fan of this, and our current hosting package includes no ads. We can easily switch from paying our current host to paying readthedocs.

@amimre, would you mind looking into these two?

amimre:
Yes, that's the plan. It's 5$ a month for open source projects (which we are): [https://about.readthedocs.com/pricing/#/community](https://about.readthedocs.com/pricing/#/community).

I'm happy to make some PRs to port the readthedocs compatibility to the older versions (v0.11 and up I guess?). I also set it up so that any future versions should be indexed and added automatically.

fkraushofer (approved):

michele-riva (changes requested):
I noticed two funny things:
1. The color of the dropdown menu for selecting the version is a bit too similar to the one of the background. Can we make it darker?
2. The references to our papers look a bit funny in the PDF. In both there's an "I ( V )" which should be without spaces (probably just `$I(V)$` in the bibtex would do.

michele-riva:
> Yes, that's the plan. It's 5$ a month for open source projects (which we are): https://about.readthedocs.com/pricing/#/community.

Sounds good. It would be good if you could take care of (i) this, (ii) stopping our current hosting package, and (iii) make sure we can still have @viperleed.org emails with whatever provider is more convenient.

> I'm happy to make some PRs to port the readthedocs compatibility to the older versions (v0.11 and up I guess?). I also set it up so that any future versions should be indexed and added automatically.

Yeah, 0.11 up seems reasonable. We could even do only the versions that we have published on pypi.

amimre:
Sure, I'll look into it tomorrow.

I did v012.1 and v012.2, but I didn't do v0.11 yet because we don't have a branch for it currently. I can make one if we want to have it.
Maybe we can discuss tomorrow morning.

amimre:
> The references to our papers look a bit funny in the PDF. In both there's an "I ( V )" which should be without spaces (probably just $I(V)$ in the bibtex would do).

I tried to fix this before, but unfortunately, I don't think there's anything that I can do about that. LaTeX formatting, including inline math, is not supported by pybtex, the processor used by `sphixcontrib-bibtex` (see also [here](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#latex-formatting-inside-bibtex-entries) and [here](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/114)).

amimre:
As discussed with @michele-riva I'll just change the formatting to plain `I(V)`.

amimre:
Also, do we want to activate link previews in the docs? See [here](https://docs.readthedocs.com/platform/stable/link-previews.html).

michele-riva:
> Also, do we want to activate link previews in the docs? See [here](https://docs.readthedocs.com/platform/stable/link-previews.html).

Looks very nice. I would activate it if possible!

doc/index.rst
```diff
@@ -144,6 +144,8 @@ project:
         License<content/appendix/license>
         content/appendix/notes_for_developers
 
+   A PDF version of this documentation is available for download in the version
+   selection menu in the bottom right corner of this page.
```
michele-riva, line 148: version-selection menu.... bottom-right corner

michele-riva (changes requested):

michele-riva:
Sorry, forgot to comment: looks good! Only one minor wording change

amimre:
Ok done. Link previews are also on now.

michele-riva (dismissed):
Good to go!

amimre:
Sorry, I just noticed a small mistake in the URL for the docs badge in the README. Could you reapprove?

michele-riva (approved):
Done!

amimre:
`viperleed.org` has moved to new servers hosted by Read the Docs.
If you’re seeing a 404 error, try clearing your browser cache or deleting any locally saved version of the site. The issue should resolve itself within 24 hours as DNS updates propagate.

PR #236, opened 2024-07-26
doc,fix:append rather than overwrite .bashrc

amimre:
Fixes the suggested syntax for adding the VIPERLEED_TENSORLEED environment variable to the local .bashrc file. It should be >> which appends, rather than > which completely overwrites the file. (I nuked my .bashrc by accident with this already...)


michele-riva (approved):
I guess this is ready for review, although you did not request it?

Looks good, anyway.

Sorry about your `.bashrc`, I was not aware of the `>>` as an 'append' option.

amimre:
Yes, thanks!

PR #238, opened 2024-07-29
fix: index error when detecting firmware folders

FlorianDoerr:
When looking for firmware folders we look through the internal folder structure in a zip file. This could have resulted in an index error for empty folders before this fix.


guilib/measure/dialogs/firmwareupgradedialog.py
```diff
@@ -384,7 +384,10 @@ def _get_archived_firmware(self, file_path):
             with ZipFile(file, mode='r') as archive:
                 # !!! We are assuming here that the first folder in the
                 # .zip file matches the name of the firmware version.
-                folder_name, *_ = archive.namelist()[0].split('/')
+                try:
+                    folder_name, *_ = archive.namelist()[0].split('/')
+                except IndexError:
```
michele-riva:
```suggestion
                except IndexError:  # Archive is empty
```

michele-riva (approved):

Issue #62, opened 2023-01-04
Improved memory management


alexmiame:
The memory requirements of TensErLEED (especially the refcalc) can become quite limiting when tackling large unit cell systems. @LutzHammer reported needing about 25 GB of memory per energy step in the refcalc for a system with ~400 atoms. This limits the possible amount of concurrent reference calculations.

@LutzHammer suggested allowing a section dependent N_CORES, i.e. different N_CORES for refcalc, deltas and search. @michele-riva suggested, it may make more sense instead to monitor current memory usage and only spawn new subprocesses if there is enough free memory (up to a max limit of N_CORES).

Perhaps we should also take another look at the TensErLEED refcalc in the long run and see if we can reduce the memory footprint somehow.

amimre:
We have found that the structure optimization section (search) can also run into memory limitations, specifically when dealing with large parameter spaces (typically many layers). On such issue was reported by @msetvin recently (see also #239). There, too many MPI processes were spawned, and when they run out of available system memory, the search would crash.

For the search, we now catch these crashes and give a warning, suggesting lowering N_CORES. As a more general solution to the problem, it would be nice to estimate and/or monitor the memory footprint of the TensErLEED subprocesses. Ideally, we would only ever spawn as many as we can afford given the system memory. As a first step, though, we could at least try to respawn processes (with a lowered N_CORES) if we detect they were killed due to memory overuse.

PR #239, opened 2024-07-30
fix: detect search out of memory

amimre:
The MPI search process may be killed by the operating system if it runs out of available system memory. This will show up in the log file as the MPI process being terminated with exit code 9.


michele-riva (changes requested):
Good to me. May I ask you to create an issue for better memory checking? We may even already have one (#62) that would need a slight update to also track this situation.

michele-riva (approved):

PR #241, opened 2024-08-05
Add dataclass utilities

michele-riva:
Adds a new `calc.lib` module that collects various bits of code useful for handling `dataclasses`.

It also adds a **new dependency**: `typing-extensions`, a backport of new `typing` features down to Python 3.7

Tests for 100% coverage.
All tests passing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
     'quicktions',
     'scipy',
     'scikit-learn',
+    'typing-extensions; python_version<"3.12"',
     'wrapt',
     ]
 keywords = [
--- a/src/viperleed/calc/lib/dataclass_utils.py
+++ b/src/viperleed/calc/lib/dataclass_utils.py
@@ -1,0 +1,23 @@
+"""Module dataclasses_utils of viperleed.calc.lib.
+
+Collects useful functions that add up on top of the dataclasses
+stdlib module.
+"""
+
+__authors__ = (
+    'Michele Riva (@michele-riva)',
+    )
+__copyright__ = 'Copyright (c) 2019-2024 ViPErLEED developers'
+__created__ = '2024-07-25'
+__license__ = 'GPLv3+'
+
+from typing import Optional
+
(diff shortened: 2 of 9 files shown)


src/viperleed/calc/lib/dataclass_utils.py
```diff
@@ -0,0 +1,227 @@
...
+else:
+    from typing import dataclass_transform
+
+
+frozen = dataclass_transform(frozen_default=True)(
+    functools.partial(dataclass, frozen=True)
+    )
+
+
+def check_types(self, init_only=False):
```
amimre, line 44: Do we need to implement this ourselves? Doesn't pylint do this anyways if there are type annotations?

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     'quicktions',
     'scipy',
     'scikit-learn',
+    'typing-extensions; python_version<"3.12"',
```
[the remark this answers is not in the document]
michele-riva: Yes, py3.7 has been supported till `v4.7.1`. Support was dropped as of `v4.8.0`. This unfortunately means that, in the long run, we may need to have a dedicated utilities module where we backport missing stuff that we need to support also `3.7`. I had thought about it for a while, and decided to add the dependency. A viable alternative would be to already make such a module, and copy over the `typing-extensions` implementations that we need from 4.7.1 to be used for 3.7.

src/viperleed/calc/lib/dataclass_utils.py
michele-riva: Unfortunately type checking for `dataclasses` in `pylint` is still pretty rudimentary. Works OK for explicit types, but not for fields that use the `dataclasses.field` function. (See the plenty issues concerning `dataclasses` at https://github.com/pylint-dev/pylint/issues

michele-riva:
I would let you judge if the conversations above are "resolved".

amimre (approved):

PR #249, opened 2024-09-03
fix: handle non-measuring controllers

FlorianDoerr:
Since non-measuring controllers would not be added to the _missing_data dictionary, decreasing their missing data count would lead to a KeyError before this fix.


michele-riva (approved):
Looks good!

PR #253, opened 2024-10-03
32 bit tisgrabber fix

FlorianDoerr:
Windows will now load the tisgrabber.dll file correctly for 32 bit python versions.


michele-riva (approved):
Good to go!

PR #254, opened 2024-10-03
fix: numpy backwards compatibility

FlorianDoerr:
Backwards compatibility fix for numpy 1.24 and below.


michele-riva (approved):
Thanks for taking care of this. Feel free to merge!

PR #266, opened 2024-10-31
More general file parser

amimre:


src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+__created__ = '2024-10-03'
+__license__ = 'GPLv3+'
+
+from abc import abstractmethod
+from collections.abc import Iterator
+from contextlib import AbstractContextManager
+from pathlib import Path
+
+
+class SettingsFileReader(AbstractContextManager, Iterator):
```
michele-riva, line 24:
```suggestion
class InputFileReader(AbstractContextManager, Iterator):
```

That's to make sure we don't confuse it with a reader for `.ini` files, also normally called "settings files" or "configuration files". Another option, considering that it reads one line at a time, would be `LineByLineFileReader`. However, if we plan to use it also to read "blocks", then the suggestion above would fit better.

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+    """A context manager that iterates the contents of a settings file
+    (e.g PARAMETERS)."""
```
michele-riva, line 26:
```suggestion
    """A context manager that iterates the contents of an input file."""
```

First line in `__doc__` should be a complete one and should fit 79 characters (incl. `"""`).

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __init__(self, filename, noisy=True):                                   # TODO: it would be nice to support passing file contents via a StringIO or similar
```
michele-riva, line 28: Generalization to a stream is simple. We could have a parent class that takes care of the stream (e.g., `InputStreamReader`). This one would derive from it. You could take a look at how we do it in `poscar`.

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            self._file_obj.close()
+        except AttributeError:
+            pass
+        return super().__exit__(exc_type, exc_value, traceback)
+
+    @abstractmethod
+    def __next__(self):
+        """Return the next understandable information in the file."""
+        raise NotImplementedError
```
michele-riva, line 61: You don't need to explicitly `raise NotImplementedError` in an `@abstractmethod`, as ABC instances (i.e., anything with an `@abstractmethod`) raise `TypeError` when instantiated.

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
+"""Module file_reader of viperleed.calc.classes.
```
michele-riva, line 1: This does not fit with where this module is placed. It currently is in `viperleed.calc.files.parameters.file_reader`. I'm not sure now where this would make more sense. My guess is in `viperleed.calc.files`, since, while it is a class, it is specifically related to reading input files.

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+
+This module is a more abstract version of a settings file reader. It is based on
```
michele-riva, line 3:
```suggestion
This module is a more abstract version of an input-file reader. It is based on
```

(See my suggestion below concerning naming of the class.)

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+@michele-riva's work on the PARAMETERS file reader, which in turn is based on
+the old readPARAMETERS from @fkraushofer.
+
+Defines context-manager, iterator classes for reading parameters from a file.
```
michele-riva, line 7:
```suggestion
Defines context-manager, iterator classes for reading information
from an input file.
```

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Close file, then return control to caller to handle exceptions."""
+        try:
+            self._file_obj.close()
+        except AttributeError:
+            pass
+        return super().__exit__(exc_type, exc_value, traceback)
+
+    @abstractmethod
+    def __next__(self):
```
michele-riva, line 59: Could we perhaps provide a simpler interface for subclasses? An example that comes to mind would be an `@abstractmethod` that takes a line as a string and produces information from it [e.g., `def _read_one_line(self, line)`]. Then we could avoid having `__next__` as an abstract one and rather produce a general-enough implementation. It could simply do the job if iterating through the file and return whatever `_read_one_line` does.

To have a general enough implementation, we would then only need to provide a good recipe on how to handle special cases. One case I can see in the PARAMETERS is skipping an 'invalid' line. It would be easy to provide a decent interface: _read_one_line could raise a dedicated exception (e.g., `ShouldSkipLineError`, defined in this module). Then a general enough implementation of `__next__` could be:

```python
    def __next__(self):
        """Return the next understandable information in the file."""
        for line in self._file_obj:
            try:
                return self._read_one_line(line)
            except ShouldSkipLineError:
                continue
        raise StopIteration

    @abstractmethod
    def _read_one_line(self, line):
        """Return understandable information from `line`.

        This method is guaranteed to be called once on each
        line read from the file.

        Parameters
        -------------
        line : str
            A single line read from `self._file_obj.`

        Returns
        --------
        info : object
            Understandable information read from `line`. This is the
            same object returned whenever this reader is iterated over.

        Raises
        -------
        ShouldSkipLineError
            If `line` does not contain any valuable information that is
            worth returning while iterating over the file.
        """
```

src/viperleed/calc/files/parameters/reader.py
```diff
@@ -27,43 +27,16 @@
...
-        return self
-
-    def __exit__(self, exc_type, exc_value, traceback):
-        """Close file, then return control to caller to handle exceptions."""
-        try:
-            self._file_obj.close()
-        except AttributeError:
-            pass
-        return super().__exit__(exc_type, exc_value, traceback)
+        super().__init__(filename, noisy)
```
michele-riva, line 39: Since you only call `super()` here, you can equally well remove this entirely.

src/viperleed/calc/files/parameters/reader.py
```diff
@@ -27,43 +27,16 @@
...
 from .errors import ParameterHasNoValueError
 from .errors import ParameterNotRecognizedError
+from .file_reader import SettingsFileReader
 from .known_parameters import did_you_mean
 from .known_parameters import from_alias
 from .utils import Assignment
 
-
-class ParametersReader(AbstractContextManager, Iterator):
+class ParametersReader(SettingsFileReader):
```
michele-riva, line 35: I cannot comment on the imports, but probably a couple can be removed.

src/viperleed/calc/files/parameters/reader.py
```diff
@@ -75,6 +48,19 @@ def __next__(self):
...
+
+    def _parse_line(self, line):
+        """Return a parameter string and an Assignment from line."""
+        pass
+
+    @staticmethod
+    def _tokenize_line(line):
+        """Split up line into tokens."""
+        pass
+
```
michele-riva, line 63: These seem duplicates, they are defined further below, with an implementation.

src/viperleed/calc/files/parameters/file_reader.py
```diff
@@ -0,0 +1,61 @@
...
+__copyright__ = 'Copyright (c) 2019-2024 ViPErLEED developers'
+__created__ = '2024-10-03'
+__license__ = 'GPLv3+'
+
+from abc import abstractmethod
+from collections.abc import Iterator
+from contextlib import AbstractContextManager
+from pathlib import Path
+
+
```
michele-riva, line 23: I can see some useful subclasses for this one. An example that comes to mind is a `DelimitedInputFileReader` that takes care of files where lines are of the form:

`<stuff on the left> <delimiter> <stuff on the right>`

Another useful subclass that may require a bit of generalization of this one: a reader for files that are split into "blocks". Then each block could of course be a reader itself. In that case, generalization to a stream would be ideal.

michele-riva (changes requested):
Some suggestions for improvements.

Also, please add some tests.

src/viperleed/calc/files/input_reader.py
```diff
@@ -0,0 +1,153 @@
...
+        Raises
+        -------
+        ShouldSkipLineError
+            If `line` does not contain any valuable information that is
+            worth returning while iterating over the file.
+        """
+        pass
+
+
+class InputStreamReader(InputReader):
```
michele-riva, line 87: It may not be apparent from reading the code that this is still an ABC. You can signal that by

```python
from abc import ABC

class InputStreamReader(ABC, InputReader):
```

michele-riva (changes requested):
Some more edits, sorry.

src/viperleed/calc/files/input_reader.py
```diff
@@ -0,0 +1,159 @@
...
+
+class ShouldSkipLineError(Exception):
+    """Exception raised when a line's contents should be skipped."""
+
+
+class InputReader(Iterator):
+    """Common base class for all input readers."""
+
+    def __init__(self, noisy=True):
+        """Initialize base class instance."""
```
michele-riva, line 40:
```suggestion
        """Initialize base-class instance.

        Parameters
        ----------
        noisy : bool, optional
            Whether the reader will emit logging messages and raise
            errors if unknown or malformed lines are encountered.

        Returns
        -------
        None.
        """
```

src/viperleed/calc/files/input_reader.py
```diff
@@ -0,0 +1,159 @@
...
+        --------
+        info : object
+            Understandable information read from `line`. This is the
+            same object returned whenever this reader is iterated over.
+
+        Raises
+        -------
+        ShouldSkipLineError
+            If `line` does not contain any valuable information that is
+            worth returning while iterating over the self.stream.
```
michele-riva, line 84:
```suggestion
            worth returning while iterating over `self.stream`.
```

michele-riva (changes requested):
Last round, I promise :)

michele-riva:
Sorry, I just realized that this was not implemented: https://github.com/viperleed/viperleed/pull/266#discussion_r1829087385. Would you mind editing?

michele-riva (approved):
Thanks!

Issue #44, opened 2022-10-19
Use of gfortran and mpifort compiler flags -std=legacy and -fallow-argument-mismatch


alexmiame:
As it currently stands, the gcc based MPI Fortran compiler will not compile the TensErLEED source code for the search because it contains implicit type conversions between REALs and INTEGERs.
This used to be standard practice in the past, but has since been deprecated for good reason (Fortran is strongly typed after all).

Modern gcc and mpifort need to be given either the -std=legacy or the -fallow-argument-mismatch flag to allow this conversion and downgrade the errors to warning. However, use of `-std=legacy` is discuraged and `-fallow-argument-mismatch` seems to cause some problems.
The better solution would be to go through the code and add either convert types explicitly or add an explicit `TYPE(*)` interface to the offending subroutines.

michele-riva:
A note about this, as it has just bit me: The `-fallow-argument-mismatch` flag is only available since `gfortran` v10. The 9.4 version that ships automatically with some Ubuntu versions will complain about it as an unknown flag. I see two ways here, aside from the **way better** solution of actually fixing the TensErLEED code:
1. support only `gfortran>=10`. In this case we should check the result of `gfortran --version` and complain. If we go this way, we have to state this clearly in the documentation. We should definitely have some more detailed instructions on how to install more recent `gfortran`, pointing people directly to https://fortran-lang.org/en/learn/os_setup/install_gfortran/#debian-based-debian-ubuntu-mint-etc. One important point that is not mentioned in the guide above is that, despite the update, `gfortran` will still be an alias for `gfortran-9`. A solution would be to re-alias the command as suggested [here](https://fortran-lang.discourse.group/t/installing-latest-gfortran/3934/4), but I'm not sure whether this may break other things.
2. support also older versions and pick the flag depending on version. It is unclear whether `gfortran` <10 would be able to compile the argument-mismatched code at all. This needs testing.

I'd like an opinion from @amimre and @fkraushofer about how to proceed here. Perhaps @amimre could try whether option 2 would work at all.

michele-riva:
A side note:

> However, use of `-std=legacy` is discuraged and `-fallow-argument-mismatch` seems to cause some problems.

Also `-fallow-argument-mismatch` is **strongly** discouraged, see https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html

michele-riva:
[...]

PR #269, opened 2024-11-06
Check Mpifort version

amimre:
As discussed in issue #44, there can be issues with the used compiler flag `-fallow-argument-mismatch`, as it does not exist for gfortran (and thus mpifort) versions <10.0. For newer versions though, compilation may error out if the flag is not provided.
This adds a check for the version and only supplies the option if a high enough version is detected.

### DIFF
--- a/src/viperleed/calc/classes/rparams/rparams.py
+++ b/src/viperleed/calc/classes/rparams/rparams.py
@@ -396,7 +396,6 @@ class Rparams:
             f'{source_tree}.'
             )
 
-
     def updateDerivedParams(self):
         """
         Checks which derivative parameters (which cannot be calculated at
@@ -419,7 +418,6 @@ class Rparams:
             tl_source = self.get_tenserleed_directory()
             self.TL_VERSION = tl_source.version
             _LOGGER.debug(f'Detected TensErLEED version {str(self.TL_VERSION)}')
-
 
         # SEARCH_CONVERGENCE:
         if self.searchConvInit['gaussian'] is None:
@@ -695,22 +693,24 @@ class Rparams:
         elif found == 'mpifort':
             # check for the mpifort version
             mpifort_version = None
-            try:
+            mpifort_call = "mpifort -Ofast"
+            try:  # Add version-dependent CLI args
                 mpifort_version = fortran_utils.get_mpifort_version()
-            except fortran_utils.CouldNotDetermineMpifortVersionError:
+            except fortran_utils.FortranCompilerError:
                 _LOGGER.warning(
-                    'mpifort version could not be determined automatically. '
-                    'mpifort versions <= 10.0 may need the '
-                    '"-fallow-argument-mismatch flag" or "-std=legacy" flag to '
-                    'compile the TenseErLEED structure search code. '
-                    'If an error occurs, please check the mpifort version and '
-                    'adapt the FORTRAN_COMP parameter as required.'
+                    "mpifort version could not be determined automatically. "
+                    "mpifort versions >= 10.0 may need the "
+                    '"-fallow-argument-mismatch" or "-std=legacy" flags to '
+                    "compile the TenseErLEED structure-search code. "
+                    "If an error occurs, please check the mpifort version and "
+                    "adapt the FORTRAN_COMP parameter as required."
                 )
-            mpifort_call = 'mpifort -Ofast'
-            if mpifort_version is not None and mpifort_version > Version('9.0'):
-                mpifort_call += ' -fallow-argument-mismatch'
-
-            self.FORTRAN_COMP_MPI = [mpifort_call, '']
+            else:
+                mpifort_call += (
+                    "" if mpifort_version < "10.0"
+                    else " -fallow-argument-mismatch"
+                )
+            self.FORTRAN_COMP_MPI = [mpifort_call, ""]
             _LOGGER.debug('Using fortran compiler: mpifort')
 
     def renormalizeDomainParams(self, config):
--- a/src/viperleed/calc/classes/rparams/rparams.py
+++ b/src/viperleed/calc/classes/rparams/rparams.py
@@ -692,7 +692,6 @@ class Rparams:
             _LOGGER.debug('Using fortran compiler: mpiifort')
         elif found == 'mpifort':
             # check for the mpifort version
-            mpifort_version = None
             mpifort_call = "mpifort -Ofast"
             try:  # Add version-dependent CLI args
                 mpifort_version = fortran_utils.get_mpifort_version()
--- a/src/viperleed/calc/classes/rparams/rparams.py
+++ b/src/viperleed/calc/classes/rparams/rparams.py
@@ -703,12 +703,11 @@ class Rparams:
                     'compile the TenseErLEED structure-search code. '
                     'If an error occurs, please check the mpifort version and '
                     'adapt the FORTRAN_COMP parameter as required.'
-                )
+                    )
             else:
-                mpifort_call += (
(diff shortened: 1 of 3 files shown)


src/viperleed/calc/lib/fortran_utils.py
```diff
@@ -11,10 +11,18 @@
 __created__ = '2024-08-26'
 __license__ = 'GPLv3+'
 
+import subprocess
+
+from .version import Version
```
michele-riva, line 16:
```suggestion
from viperleed.calc.lib.version import Version
```

src/viperleed/calc/lib/fortran_utils.py
```diff
@@ -11,10 +11,18 @@
...
 
 _FORTRAN_LINE_LENGTH = 72  # FORTRAN line-length limit
 _F77_CONTINUATION_POS = 6  # Column of the continuation character
 
+class MpifortNotFoundError(Exception):
```
michele-riva, line 21: I would perhaps make these exceptions not related to `mpifort`, since they may be used later on for other compilers. I'd suggest: `CompilerNotFoundError` and `NoCompilerVersionFoundError`. I would make them both a subclass of a more general `FortranCompilerError` exception.

michele-riva (changes requested):
Some suggestions to be implemented.

If it turns out that we cannot support mpifort < 10 (see my comments below for suggestions on some checks), I would then also update the documentation.

src/viperleed/calc/classes/rparams/rparams.py
```diff
@@ -692,10 +693,24 @@ def getFortranMpiComp(self, comp='auto'):
...
+                    '"-fallow-argument-mismatch flag" or "-std=legacy" flag to '
+                    'compile the TenseErLEED structure search code. '
+                    'If an error occurs, please check the mpifort version and '
+                    'adapt the FORTRAN_COMP parameter as required.'
+                )
+            mpifort_call = 'mpifort -Ofast'
+            if mpifort_version is not None and mpifort_version > Version('9.0'):
+                mpifort_call += ' -fallow-argument-mismatch'
+
+            self.FORTRAN_COMP_MPI = [mpifort_call, '']
```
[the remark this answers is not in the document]
amimre: I tried it now with the earlier version of gfortran and mpifort. Results are as expected:
While both `-Wargument-mismatch` and `-std=legacy` exist for earlier versions, they are not necessary. Argument mismatch was not an error by default prior to 10.0, so it compiles fine without either.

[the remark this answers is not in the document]
michele-riva: Ah I see. Then no need to have any extra CLI argument here.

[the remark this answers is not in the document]
michele-riva: Would you mind implementing the suggested changes? It's mostly some stylistic ones (and some fixes on wording errors).

src/viperleed/calc/classes/rparams/rparams.py
```diff
@@ -692,10 +693,24 @@ def getFortranMpiComp(self, comp='auto'):
             self.FORTRAN_COMP_MPI = ['mpiifort -Ofast', '']
             _LOGGER.debug('Using fortran compiler: mpiifort')
         elif found == 'mpifort':
-            self.FORTRAN_COMP_MPI = [
-                'mpifort -Ofast -no-pie -fallow-argument-mismatch',
-                ''
-                ]
+            # check for the mpifort version
```
michele-riva, line 696: Please implement changes I suggested in previous round on this part.

src/viperleed/calc/lib/fortran_utils.py
```diff
@@ -5,17 +5,64 @@
...
+    'Alexander M. Imre (@amimre)',
     'Michele Riva (@michele-riva)',
     )
 __copyright__ = 'Copyright (c) 2019-2024 ViPErLEED developers'
 __created__ = '2024-08-26'
 __license__ = 'GPLv3+'
 
+import re
+import subprocess
+import shutil
```
michele-riva, line 17: swap for alphabetic order:
```suggestion
import shutil
import subprocess
```

michele-riva (changes requested):
Some small changes remaining. Please make sure to include the changes I suggested earlier for rparams.

src/viperleed/calc/classes/rparams/rparams.py
```diff
@@ -692,10 +691,26 @@ def getFortranMpiComp(self, comp='auto'):
...
+            mpifort_version = None
+            mpifort_call = "mpifort -Ofast"
```
michele-riva, line 696: We use single quotation marks everywhere else in this function. It would be good to stick to them also here. I think we decided a while back that single quotes would be our standard. The only exception is the triple quotes in docstrings. Portions of the code from Florian that we haven't touched much still have double quotes, but I would not introduce more with new code, and rather slowly migrate double to single ones in parts of the code that we progressively touch.

src/viperleed/calc/classes/rparams/rparams.py
```diff
@@ -692,10 +691,26 @@ def getFortranMpiComp(self, comp='auto'):
...
+                mpifort_version = fortran_utils.get_mpifort_version()
+            except fortran_utils.FortranCompilerError:
+                _LOGGER.warning(
+                    "mpifort version could not be determined automatically. "
+                    "mpifort versions >= 10.0 may need the "
+                    '"-fallow-argument-mismatch" or "-std=legacy" flags to '
+                    "compile the TenseErLEED structure-search code. "
+                    "If an error occurs, please check the mpifort version and "
+                    "adapt the FORTRAN_COMP parameter as required."
+                )
```
michele-riva, line 707: I think we decided to have closing parentheses aligned at the same column as the previous argument block.

```suggestion
                    "adapt the FORTRAN_COMP parameter as required."
                    )
```

(multiple occurrences)

michele-riva (changes requested):
Only small style changes left

src/viperleed/calc/classes/rparams/rparams.py
amimre: Yeah, this is why I usually use a code formatter for my own stuff, so I don't have to bother doing this (and indent, line breaks etc.) manually...

amimre:
Alright, implemented.
Tbh, though: If we're going to be this strict on style edits every time, we should really sit down and decide on a style that can be enforced with an auto-formatter (such as autopep8, black etc.).
Otherwise, it's really a pain to constantly go back and forth in every PR.
@fkraushofer what do you think?

src/viperleed/calc/classes/rparams/rparams.py
```diff
@@ -692,10 +691,24 @@ def getFortranMpiComp(self, comp='auto'):
...
+                    'mpifort versions >= 10.0 may need the '
+                    '"-fallow-argument-mismatch" or "-std=legacy" flags to '
+                    'compile the TenseErLEED structure-search code. '
+                    'If an error occurs, please check the mpifort version and '
+                    'adapt the FORTRAN_COMP parameter as required.'
+                    )
+            else:
+                mpifort_call += ('' if mpifort_version < '10.0'
+                                 else ' -fallow-argument-mismatch'
+                                )
```
michele-riva, line 710:
```suggestion
                                 else ' -fallow-argument-mismatch')
```

Here, since the parenthesized part is not on a new line, the closing parenthesis goes on the last line.

michele-riva (changes requested):
Ok this is really the last one...

michele-riva (approved):
Looks good!

PR #270, opened 2024-11-07
Use all hardware threads with Open MPI

michele-riva:
This PR addresses a problem that surfaced in the context of #44, specifically https://github.com/viperleed/viperleed/issues/44#issuecomment-2457638938.

This PR implements the suggestion as in https://github.com/viperleed/viperleed/issues/44#issuecomment-2459148042.

I originally intended to honor the suggestion by @amimre in https://github.com/viperleed/viperleed/issues/44#issuecomment-2459415952. However the two of us have tested a bit yesterday the addition of the `--use-hwthread-cpus` flag to `mpirun` and came to the conclusion that it is safe to always use, irrespective of whether `N_CORES` was specified or auto-detected.

The concern voiced in https://github.com/viperleed/viperleed/issues/44#issuecomment-2459415952 regarded especially the usage in an HPC environment. Testing on the [VSC-4](https://vsc.ac.at/home/) showed that, on compute nodes that are managed by SLURM, the presence of the `--use-hwthread-cpus` flag has no impact on the number of slots that `mpirun` has access to. I.e., both

```
mpirun --display-map -n 1 true
mpirun --display-map --use-hwthread-cpus -n 1 true
```

give the same number of slots. 

I'm happy to come up with a bit more detailed solution of adding the flag only if `N_CORES` was detected automatically (and error out cleanly otherwise) in case @fkraushofer and @amimre still think it would be better.

Notice that the flag does not imply that Open MPI will do any oversubscription. It will use as many threads as are available from either the nominal count from the CPU manufacturer, or whatever is limited by a task scheduler (like SLURM).

### DIFF
--- a/tests/calc/files/parameters/test_checker.py
+++ b/tests/calc/files/parameters/test_checker.py
@@ -7,6 +7,8 @@ __copyright__ = 'Copyright (c) 2019-2024 ViPErLEED developers'
 __created__ = '2023-10-25'
 __license__ = 'GPLv3+'
 
+from unittest.mock import MagicMock
+
 import pytest
 from pytest_cases import fixture, parametrize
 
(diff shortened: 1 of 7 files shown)


amimre:
I guess you probably want to rebase/merge into 0.12.2, otherwise you merge all a bunch of other changes into master.

michele-riva:
> I guess you probably want to rebase/merge into 0.12.2, otherwise you merge a bunch of other changes into master.

Thanks, I didn't notice.

I'll check one more thing (i.e., what happens if N_CORES > maximum? --> we should probably complain with some understandable error), then I'll ask for a review.

amimre:
> My investigation about having a large `N_CORES` suggests that Intel oneAPI (v2024.2) does automatically oversubscribe irrespective of the number of threads. Open MPI, instead fails as in [#44 (comment)](https://github.com/viperleed/viperleed/issues/44#issuecomment-2457638938).
> 
> I will cleanly error out for the Open MPI case. How do we want to behave for the Intel one? Should we generally emit a warning if we find out the users are requesting a number of cores that is larger than the available ones?

For Intel, I would continue with the user requested number of processes – possibly with a warning if it's easy to implement.

michele-riva:
Alright. Yes, I had no intentions of fiddling with user needs. I would emit the warning irrespective of which compiler was chosen, as it is probably non a good idea to do a lot of oversubscribing anyway.

tests/calc/files/parameters/test_checker.py
```diff
@@ -63,3 +67,83 @@ def test_missing_parameters(self, rpars_kwargs, checker, rpars_with_attrs):
         rpars = rpars_with_attrs(**rpars_kwargs)
         with not_raises(Exception):
             checker.check_parameter_conflicts(rpars)
+
+
+class TestFortraCompUpdated:
```
amimre, line 72: Typo: Fortran

tests/calc/files/parameters/test_checker.py
```diff
@@ -7,6 +7,8 @@
 __created__ = '2023-10-25'
 __license__ = 'GPLv3+'
 
+from unittest.mock import MagicMock
```
amimre, line 10: It doesn't really matter as they are compatible, but is there a way to do this in pytest?

amimre (commented):
Looks good, one small typo

tests/calc/files/parameters/test_checker.py
michele-riva: 600e9b5f5fe22c3ef1fc22b97b021433e763439a

amimre (approved):

tests/calc/files/parameters/test_checker.py
michele-riva: Thanks for the input. Indeed, there is [pytest-mock](https://pytest-mock.readthedocs.io/en/stable/index.html), which makes mocking way nicer. I've added it as an additional test dependency and updated the tests. 1bc4e3932c70da5e076ff045ee6122224bd0a855

amimre (approved):

Issue #174, opened 2024-02-18
Internally use same frame for slab atoms and unit cell


michele-riva:
In the current implementation we use two conflicting reference frames for the unit-cell vectors and for (the Cartesian position of) Atom objects:
* unit cell: **right-handed** system, with _x_, _y_ in the surface plane, _z_ going **from the bulk towards vacuum**. `z=0` is the **bottom** of the slab (i.e., in the bulk).
* atoms: **left-handed** system, with _x_, _y_ in the surface plane [each parallel to those of the unit cell], _z_ going **from vacuum into the solid**. The **topmost** atom (i.e., at the surface) has `.cartpos[2] == 0`.

The **fractional** coordinates of the atoms are expressed in the unit-cell frame. This means that the **bottom** atom has `.pos[2] == 0`.

This convention is confusing, and makes some bits of code hard to follow: atom Cartesian coordinates are most often needed in the same frame as the one of the unit cell.

The original reason for the choice is that the coordinate frame used for atoms is the one common in LEED programs. However, we need Cartesian coordinates in this frame essentially only when writing input files for TensErLEED.

We could simplify a bunch of code if we were to use the **unit cell** frame for all. We would essentially only need a `Atom` `@property` that returns the Cartesian coordinates with the LEED convention. Something like

```python
class Atom:
    ...
    @property
    def leed_cartpos(self):
        cartpos = self.cartpos.copy()
        cartpos[2] = self.slab.topat_ori_z - cartpos[2]
        return cartpos
```

I had a bit of discussion a while back with @fkraushofer on what to watch out for, but I feel it would be better if we put notes down here. The tricky bits have to do with storing `slab.topat_ori_z` **at the right time**. If I recall correctly, this must be the position of the topmost atom **during the most recent reference calculation**. This looks like a job for `oriState` (which I'd anyway rename to match better its meaning).

To do: Keep also in mind #107 when dealing with this.

It would be great if @fkraushofer could point at some possible pitfalls in doing this, as he's dealt with the mess the most.

PR #271, opened 2024-11-13
feat: add --force option to modify_vacuum utility

amimre:
As pointed out by @andr3ac, it can be useful for people to be able to fully remove vacuum when building slabs.


michele-riva (changes requested):
Some small suggestions.

Please also make sure to add tests for full coverage. I currently see that some parts seem not to be covered (even before this PR): exceptions at L66 and L74, and the `parse_cli_args` method.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -24,7 +24,7 @@
 logger = logging.getLogger(__name__)
 
 
-def modify_vacuum(slab, vacuum_gap_size, absolute=False):
+def modify_vacuum(slab, vacuum_gap_size, absolute=False, force=False):
```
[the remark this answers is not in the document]
amimre: I would keep the name force. `-f`/`--force` is the canonical flag to suppress an error message/do something despite a warning for CLI programs.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -116,6 +119,12 @@ def add_parser_arguments(self, parser):
                   'or remove from the slab.'),
             action='store_true'
             )
+        parser.add_argument(
+            '-f', '--force',
```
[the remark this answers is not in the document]
amimre: See above; also it's not just zero but we would errour out for any vacuum gap < 5A.

src/viperleed/utilities/poscar/modify_vacuum.py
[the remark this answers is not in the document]
michele-riva: Ok, perhaps we can keep the name for the CLI flag, but I would make the kwarg name more explicit. An alternative name: `accept_small_gap`.

src/viperleed/utilities/poscar/modify_vacuum.py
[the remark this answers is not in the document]
michele-riva: Ah, yes, you're right, it's not just zero.

src/viperleed/utilities/poscar/modify_vacuum.py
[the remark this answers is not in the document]
amimre: Ok, sounds good.

tests/calc/utilities/poscar/test_modify_vacuum.py
```diff
@@ -54,12 +58,36 @@ def test_modify_vacuum_gap_absolute(self, test_slab, vacuum_gap_size):
...
+
+    def test_parse_cli_args_absolute_negative_gap(self):
+        """Test that parse_cli_args raises for a negative absolute gap."""
+        parser = ModifyVacuumCLI()
+        with pytest.raises(SystemExit):
+            parser.parse_cli_args(["-1.0", "--absolute"])
+
+    @parametrize('vacuum_gap_size', [-1.0, -5.0])
+    @infoless
+    def test_negative_gap_raises(self, test_slab, vacuum_gap_size):
```
michele-riva, line 79: Looks like lines 79-80 are still not covered. Would be good to cross check correctness, as there is a suspicous use of `atom.cartpos[2]` that is zero for the topmost atom, and largest for the bottommost one #174.

Notice that `BaseSlab` has a `.top_atom` property (but its `.cartpos[2]` is still affected by #174).

michele-riva (changes requested):
Some more small changes, plus on logical check for a portion uncovered by tests.

tests/calc/utilities/poscar/test_modify_vacuum.py
amimre: Thinking about it, I think the check for the vacuum gap on top is actually useless, since we shift the slab to the bottom of the cell anyhow. I would remove it and only check that the gap is >=0.

michele-riva: Sure, if this does not break the rest, feel free to remove it. However, we seem to be moving the vacuum to the top only if there is more than 5AA. Check whether removing this portion works also for slabs with smaller vacuum gaps.

You could even try to entirely remove the check for negative gap, as `check_vacuum_gap` may do the right thing. Should be tested though.

amimre: The check for >=0 has to stay, as still has to be an error even with allow_small_vacuum_gap. But I will check the rest.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -127,10 +143,15 @@ def parse_cli_args(self, args):
 
     def process_slab(self, slab, args):
         """Return a new slab with modified vacuum."""
+        vacuum_gap_info = VacuumGapInfo(
+            size=args.vacuum,
+            absolute=args.absolute,
+            accept_small_gap=args.force,
```
michele-riva, line 149: We don't seem to be testing the `accept_small_gap == True` case. Also, `parse_cli_args` is only tested for the "failing" case.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -55,42 +72,35 @@ def modify_vacuum(slab, vacuum_gap_size, absolute=False):
...
     new_c_vector = (
         processed_slab.c_vector
         / np.linalg.norm(processed_slab.c_vector)
         * (vacuum_gap_size + slab_thickness)
     )
-    print(f'New c vector: {new_c_vector}')
-    print(f'Old c vector: {processed_slab.c_vector}')
-
     processed_slab.c_vector[:] = new_c_vector
     processed_slab.collapse_cartesian_coordinates()
```
michele-riva, line 92: Now that I think about it, I feel like you probably need an `update_origin=True` in here, since atoms do move along _z_ as a result of the resizing of the **c** vector.

michele-riva (changes requested):
Sorry, few more changes.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -12,31 +12,52 @@
...
+from dataclasses import dataclass
 import logging
 
 import numpy as np
 
-from viperleed.calc.classes.slab.errors import (NotEnoughVacuumError,
+from viperleed.calc.classes.slab.errors import (
+                                                NotEnoughVacuumError,
+                                                VacuumError,
                                                 WrongVacuumPositionError)
```
michele-riva, line 23:
```suggestion
from viperleed.calc.classes.slab.errors import NotEnoughVacuumError
from viperleed.calc.classes.slab.errors import VacuumError
from viperleed.calc.classes.slab.errors import WrongVacuumPositionError
```

Better keep one explicit import per line. Makes it easier to add/remove imports later, as (i) the commit involves only a single line, and (ii) you don't risk forgetting a trailing comma.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -55,42 +76,38 @@ def modify_vacuum(slab, vacuum_gap_size, absolute=False):
...
         processed_slab.check_vacuum_gap()
-    except NotEnoughVacuumError:
-        raise RuntimeError('The resulting vacuum gap would be too small.')
-    except WrongVacuumPositionError:
-        raise RuntimeError('Cannot modify the vaccum gap as requested. Check '
-                           'that there already is a vacuum gap in the POSCAR.')
+    except NotEnoughVacuumError as err:
+        if not vacuum_gap_info.accept_small_gap:
+            raise NotEnoughVacuumError(
+                'The resulting vacuum gap would be too small.', None) from err
```
michele-riva, line 105:
```suggestion
            raise NotEnoughVacuumError(
                'The resulting vacuum gap would be too small.',
                None,
                ) from err
```

Again, when using the form "call(\n<arg>,...)" make sure to have one argument per line, aligned at the same column, and the closing parenthesis on a line by itself, also aligned at the same column.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -55,42 +76,38 @@ def modify_vacuum(slab, vacuum_gap_size, absolute=False):
...
+    except WrongVacuumPositionError as err:
+        if not vacuum_gap_info.accept_small_gap:
+            raise WrongVacuumPositionError(
+                'Cannot modify the vacuum gap as requested. Check that there '
+                'already is a vacuum gap in the POSCAR.', None) from err
```
michele-riva, line 110:
```suggestion
                'Cannot modify the vacuum gap as requested. Check '
                'that there already is a vacuum gap in the POSCAR.',
                None,
                ) from err
```

Ditto. Also, for strings (and comments) that need to span multiple lines, it looks better if they're wrapped to more-or-less the same width.

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -12,31 +12,52 @@
...
+        gap smaller than the minimum value for a viperleed.calc
+        run, as long as it is non-negative. Default is False.
+    """
+
+    size: float
+    absolute: bool = False
+    accept_small_gap: bool = False
+
+
+def modify_vacuum(slab, vacuum_gap_info):
```
michele-riva, line 52: I cannot comment on L69 below, but please update the documentation of what this function raises and when.

tests/calc/utilities/poscar/test_modify_vacuum.py
```diff
@@ -9,25 +9,29 @@
...
 import numpy as np
 import pytest
-import pytest_cases
+from pytest_cases import parametrize
+from pytest_cases import parametrize_with_cases
 
-from viperleed.utilities.poscar.modify_vacuum import ModifyVacuumCLI
+from viperleed.calc.classes.slab.errors import VacuumError
 from viperleed.calc.classes.slab.surface_slab import _MIN_VACUUM
+from viperleed.utilities.poscar import modify_vacuum
```
michele-riva, line 17:
```suggestion
from viperleed.utilities.poscar.modify_vacuum import modify_vacuum
```

Perhaps nicer, it avoids the double `modify_vacuum.modify_vacuum`. If you go this way, also re-sort the imports so this comes after exceptions (capitals come earlier in ASCII).

tests/calc/utilities/poscar/test_modify_vacuum.py
```diff
@@ -54,12 +58,56 @@ def test_modify_vacuum_gap_absolute(self, test_slab, vacuum_gap_size):
...
+            parser.parse_cli_args(['-1.0', '--absolute'])
+
+    @parametrize('vacuum_gap_size', [-1.0, -5.0])
+    @infoless
+    def test_negative_gap_raises(self, test_slab, vacuum_gap_size):
+        """Check complaints when the vacuum gap size is negative."""
+        slab, *_ = test_slab
+        vacuum_gap_info = VacuumGapInfo(size=vacuum_gap_size, absolute=True)
+        with pytest.raises(VacuumError):
+            modified_slab = modify_vacuum.modify_vacuum(slab, vacuum_gap_info)
```
michele-riva, line 84: Adapt depending on my comment at L17 (multiple occurrences).

tests/calc/utilities/poscar/test_modify_vacuum.py
```diff
@@ -54,12 +58,56 @@ def test_modify_vacuum_gap_absolute(self, test_slab, vacuum_gap_size):
...
-    def test_modify_vacuum_gap_too_small(self, test_slab, vacuum_gap_size):
+    def test_gap_too_small(self, test_slab, vacuum_gap_size):
         """Test the ModifyVacuumCLI class with absolute vacuum gap size."""
         parser = ModifyVacuumCLI().parser
         slab, *_ = test_slab
         args = parser.parse_args([str(vacuum_gap_size), '-a'])
         with pytest.raises(SystemExit):
             modified_slab = ModifyVacuumCLI().process_slab(slab, args)
+
+    def test_parse_cli_args_absolute_negative_gap(self):
```
michele-riva, line 71: The non-failing case for `parse_cli_args` is still untested (absolute, with positive vacuum; relative vacuum with both positive and negative values). Could you please add some tests?

tests/calc/utilities/poscar/test_modify_vacuum.py
```diff
@@ -54,12 +58,56 @@ def test_modify_vacuum_gap_absolute(self, test_slab, vacuum_gap_size):
...
+        # so we shouldn't test for the exact gap size
+        modified_slab = modify_vacuum.modify_vacuum(slab, gap)
+
+    @infoless
+    def test_zero_gap_raises(self, test_slab):
+        """Check complaints when the vacuum gap size is negative."""
+        slab, *_ = test_slab
+        vacuum_gap_info = VacuumGapInfo(size=0., absolute=True)
+        if slab.thickness <= 1e-8:  # monolayer; would lead to zero volume cell
+            return
```
michele-riva, line 111:
```suggestion
        if slab.thickness <= 1e-8:
            pytest.skip('Single layer; would lead to zero volume cell')
```

Ditto

tests/calc/utilities/poscar/test_modify_vacuum.py
```diff
@@ -54,12 +58,56 @@ def test_modify_vacuum_gap_absolute(self, test_slab, vacuum_gap_size):
...
+        gap = VacuumGapInfo(size=0.5, absolute=True, accept_small_gap=False)
+        with pytest.raises(VacuumError):
+            modified_slab = modify_vacuum.modify_vacuum(slab, gap)
+
+    @infoless
+    def test_accept_small_gap(self, test_slab):
+        """Test that NotEnoughVacuumError is raised correctly."""
+        slab, *_ = test_slab
+        if slab.thickness <= 1e-8:  # monolayer; would lead to zero volume cell
+            return
```
michele-riva, line 99:
```suggestion
        if slab.thickness <= 1e-8:
            pytest.skip('Single layer; would lead to zero volume cell')
```

Nicer so that we don't get a "passing" test but a "skipped" one. Then one can more easily find out why it was skipped, without going into the code (with `pytest -rS tests/...`)

michele-riva (changes requested):
Some more suggestions, hopefully last round!

tests/calc/utilities/poscar/test_modify_vacuum.py
michele-riva: Already done in b96e0fdab7ccf51d298a6d16f99edb2aa87b04a8

tests/calc/utilities/poscar/test_modify_vacuum.py
michele-riva: Already done in b96e0fdab7ccf51d298a6d16f99edb2aa87b04a8

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -55,42 +76,38 @@ def modify_vacuum(slab, vacuum_gap_size, absolute=False):
...
+    processed_slab.collapse_cartesian_coordinates(update_origin=True)
+
     try:
         processed_slab.check_vacuum_gap()
-    except NotEnoughVacuumError:
-        raise RuntimeError('The resulting vacuum gap would be too small.')
-    except WrongVacuumPositionError:
-        raise RuntimeError('Cannot modify the vaccum gap as requested. Check '
-                           'that there already is a vacuum gap in the POSCAR.')
+    except NotEnoughVacuumError as err:
```
michele-riva, line 102: Sorry, one more comment: please use `exc` instead of `err` in `except ... as ...`.

michele-riva (changes requested):
(one more comment since last review)

src/viperleed/utilities/poscar/modify_vacuum.py
amimre: Fine by me, but I think that goes against the pylint suggestions that you are usually so strict on.

michele-riva: I'm not sure I understand which suggestion does this go against:
- [ungrouped-imports](https://pylint.readthedocs.io/en/stable/user_guide/messages/convention/ungrouped-imports.html) only says one should have all `import` statements from the same module in consecutive lines
- [PEP-8](https://peps.python.org/pep-0008/#imports) does not seem to state anything about deciding to have one import per line.

Could you please share a reference to where this is stated to be an anti-pattern?

src/viperleed/utilities/poscar/modify_vacuum.py
```diff
@@ -45,8 +66,19 @@ def modify_vacuum(slab, vacuum_gap_size, absolute=False):
 
     Raises
     ------
-    RuntimeError
-        If the resulting vacuum gap size is negative.
+    NotEnoughVacuumError
+        If the resulting vacuum gap size is negative or too small and 
+        `accept_small_gap` is not enabled in `vacuum_gap_info`.
```
michele-riva, line 71:
```suggestion
        If the resulting vacuum-gap size is negative, or if
        `accept_small_gap` is not enabled in `vacuum_gap_info`
        and the resulting vacuum-gap size is smaller than 5 A.
```

Small rewording suggestion.

michele-riva (changes requested):
One last very small wording change.

src/viperleed/utilities/poscar/modify_vacuum.py
amimre: As far as I understand, PEP328: https://peps.python.org/pep-0328/

michele-riva: That looks like the PR that made parantheses possible in imports, which I think are still acceptable when one of the import statements goes beyond the 79 char limit. In that case, we have adopted the following style:

```python
from long_module_name.with.a_lot.of_submodules import (
    some_short_name,
    some_very_long_name_that_would_not_fit_pep8,
    )
```

Notice that, like in the case of this comment, there is one imported quantity per line.

michele-riva:
Please re-request review once you're done with the edits

michele-riva (approved):
Ready to merge. Thanks for putting up with my usual pickiness.

PR #276, opened 2024-11-20
Ruff linter formatter

amimre:


fkraushofer:
Very good idea. I also feel that there is far too much time spent on style-perfectionism. While spending _some_ time on making sure the formatting is consistent is certainly good for readability, I think it's easy to lose track of the fact that readability is the primary goal, and achieved much sooner than perfection. Everything else is optional. One should not let this take a too large fraction of the total available programming time imo; there's always better things to do.

I don't care about the specific rules at all, as long as they are clear.

michele-riva:
Good idea. Less work for me too.

amimre:
I merged the current version of this into the dev/try-me branch. Feel free to try it out and let me know how it works.
I would refrain from auto-formatting existing files, though, so we don't end up with a million merge conflicts.

fkraushofer (dismissed):
Looks very good at first glance. I tested it on a few existing files and liked the results. I see no issue with merging this as-is, using it for all new files from now on, and if we don't like some of the formatting we will find out and can discuss it as we merge new code in.

The only thing that tripped me is that it removed a bunch of line breaks, which then gave me warnings in the built-in Spyder linter. It turns out that the default line-length limit in Spyder, which I was using for a lot of my (especially older) files is 79, while here it is 80. I have no real preference, I can also simply change my settings in Spyder; if we currently have two different standards (some files written with limit 79, some with limit 80), then I think it makes more sense to settle on 80 as the new global limit.

PR #286, opened 2024-12-03
Fix allow overwriting POSCAR file in poscar utils

### LINKED ISSUE #285: poscar utilities fail when output and input file are the same
I just realized that the following fails:
`viperleed poscar sort_by_z --infile POSCAR --outfile POSCAR`
Tested for a valid POSCAR (the utility works if the output goes to a different file); error message is:
```
Invalid unit-cell vectors. Expected shape (3, 3), found (0,) instead
usage: sort_by_z [-h] [--version] [--infile INFILE] [--outfile OUTFILE] [-v]
                 [-r]
sort_by_z: error: Failed to read POSCAR. Stopping. Info: Invalid unit-cell vectors. Expected shape (3, 3), found (0,) instead
```
The POSCAR file is completely empty afterwards.

It seems a safe bet that this is because the outfile is opened (clearing it) before reading (of the same file) is done. The easy fix would be to read in the entire slab, then close the infile, do the processing in memory, and only then open the outfile for writing. I'm not sure if it's possible for io-stream-based calls (`viperleed poscar sort_by_z <POSCAR >POSCAR`), but I feel there should be at least some option that lets you make in-place modifications on a file, without having to copy and rename. At the very least, we should try to catch this in a way that doesn't result in effectively deleting the input file.

amimre:
Fixes #285.

As pointed out by @fkraushofer, there is currently an issue when trying to read and write to the same file using the poscar utilities.
This was because the FileType class from the argparse library opens the file upon parsing the argument. When the same file is selected for reading and writing, it would silently open it twice and erase the contents for the write operation. This is a know problematic behaviour (see https://github.com/python/cpython/issues/58032).
To circumvent this, argpase suggests using pathlib Paths and manual handling, which I do in this fix.

I also added tests that try to read and write from the same file to avoid regression.

### DIFF
--- a/src/viperleed/utilities/poscar/base.py
+++ b/src/viperleed/utilities/poscar/base.py
@@ -13,8 +13,8 @@ __created__ = '2024-03-27'
 __license__ = 'GPLv3+'
 
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
 from copy import deepcopy
-from io import TextIOWrapper
 import logging
 import pathlib
 import sys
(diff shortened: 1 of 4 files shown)


fkraushofer (approved):
Looks good!

Issue #43, opened 2022-10-13
Create Base Class for Compile (& Run) Tasks


alexmiame:
It would be nice to have a CompileTask and RunTask base class as these are basically redefined in the ref-calc, search, and deltas. Not urgent.

PR #298, opened 2025-01-30
Support paths relative to origin in DOMAINS

### LINKED ISSUE #297: Relative DOMAIN paths
While we currently advertise that `DOMAIN` can be specified as relative paths, this seems to be currently broken. The reason is that, when calling `parameters.interpret` in `run_calc` here https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/run.py#L147-L160 we are already inside the work directory. In fact, we move there in both the CLI and `from_ase` before `run_calc` is called: https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/cli.py#L62-L77 https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/from_ase.py#L285-L303

(In the `from_ase` case we seem to even call `.interpret` before `run_calc`, but this also happens after moving to work.)

I would like some input from @fkraushofer and @amimre on how to best tackle this. At this point, I'm not quite sure when we pull inputs for the various domains into work. I could not seem to find this in the code.

Notice that a similar issue arises during initialization: https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/sections/initialization.py#L410-L479

At that point we're also in work and paths are relative to that directory.

I'm not quite sure what the best course of action is here. It sounds like we should move over the folders/zip for domains way earlier, before `run_calc` executes. An alternative: disallow relative paths in `DOMAIN` (and document this). Thoughts?

michele-riva:
Fixes #297


michele-riva:
@fkraushofer: I started working on this, and thought I would factor out all the paths that we have as attributes of `Rparams` to simplify a bit the data structure. This lead me to look a bit closer at where we currently use `rp.workdir` (now renamed to `rp.paths.work`). I noticed that, relative to the original code that you originally wrote, we now are using this in several spots here and there (although not in a consistent manner). I then realized that this may be a potential issues for domain calculations (as mentioned in point 2 of https://github.com/viperleed/viperleed/pull/198#issuecomment-2474020727), as `dp.rp.paths.work` is set to `rp.paths.work`, and we usually pass along `dp.rp` when running calculations for a subdomain.

So the questions are:
1. Why do we store the **main** work directory in `dp.rp` rather than the one of the specific domain? This happens here (`home` is `Path.cwd()`): https://github.com/viperleed/viperleed/blob/b3a92678c08d4da2d8ae796b5ec628401378c728/src/viperleed/calc/sections/initialization.py#L481-L496 The original commit that introduced this was ed227664bd9c41adc3469d3371de91654cad48ff. Notice, however, that the original line that used `rp.workdir` (was used to fetch the TensErLEED source) is now completely absent.
2. Alternative to the above: do you feel that the various uses of `rp.paths.work` (see "Files changed" tab) should be revisited to make sure we're not using the wrong path? Could we have a look at it together?

fkraushofer:
> The original commit that introduced this was https://github.com/viperleed/viperleed/commit/ed227664bd9c41adc3469d3371de91654cad48ff. Notice, however, that the original line that used rp.workdir (was used to fetch the TensErLEED source) is now completely absent.

Yes, the main role of rp.workdir here was that it specified where the "source" files (TensErLEED and psgen stuff) could be found. At the time, all of that was not fetched from a consistent source directory, but copied to the work directory by a `job.sh` script. I believe that this also answers 1., i.e. the domains primarily needed to "know" where to find the TensErLEED fortran files, and this was (somewhat confusingly) named the "work" path (at that time, the "work" directory was understood as the one created and populated by the `job.sh` script). That was obviously the same path for all the domains. However, this is now all completely obsolete, as we're handling work and source directories in a much cleaner manner overall. Based on https://github.com/viperleed/viperleed/commit/ed227664bd9c41adc3469d3371de91654cad48ff, I'm not even sure that the rp.workdir is even required any more.

(One small side note, just because I happened to look at it right now: If you're now revisiting all the paths, please note that the "shortpath" (relative) in psgen.py needs to stay short, as EEASiSSS reads a limited number of characters)

> Alternative to the above: do you feel that the various uses of rp.paths.work (see "Files changed" tab) should be revisited to make sure we're not using the wrong path? Could we have a look at it together?

I'm honestly not sure whether this couldn't be factored out completely - but it's a bit complex to say at a glance, so yeah, maybe let's go through it together (not today though, sorry)

michele-riva:
> Yes, the main role of rp.workdir here was that it specified where the "source" files (TensErLEED and psgen stuff) could be found. At the time, all of that was not fetched from a consistent source directory, but copied to the work directory by a `job.sh` script. I believe that this also answers 1., i.e. the domains primarily needed to "know" where to find the TensErLEED fortran files, and this was (somewhat confusingly) named the "work" path (at that time, the "work" directory was understood as the one created and populated by the `job.sh` script). That was obviously the same path for all the domains. However, this is now all completely obsolete, as we're handling work and source directories in a much cleaner manner overall. Based on [ed22766](https://github.com/viperleed/viperleed/commit/ed227664bd9c41adc3469d3371de91654cad48ff), I'm not even sure that the rp.workdir is even required any more.

> I'm honestly not sure whether this couldn't be factored out completely - but it's a bit complex to say at a glance, so yeah, maybe let's go through it together (not today though, sorry)

Many thanks for the insights. Indeed, this solves the question. I personally am very much in favor of removing `rp.workdir` (now `rp.paths.work`) entirely, as this seem to cause more troubles than anything else. In most of the code that I've looked at it seems entirely redundant. There is only one spot (`refcalc`) where I'm a bit fuzzy about it. Let me know when you would have some time to sit down on Zoom and have a look through the code. Perhaps I would open a dedicated PR for that, so can I finish up this one in the meantime.
 
> (One small side note, just because I happened to look at it right now: If you're now revisiting all the paths, please note that the "shortpath" (relative) in psgen.py needs to stay short, as EEASiSSS reads a limited number of characters)

Agreed. I also saw that there were some potentially incorrect edits in `psgen`. I plan to have a go at it too.

michele-riva:
@fkraushofer: would you mind taking a quick look at it?

One question: I find the `rp.paths.source` name to be a bit confusing. It may be misinterpreted as the current `home` (i.e., where `calc` started -- name is open for discussion). Perhaps I renaming it to `rp.paths.tensorleed` would be better?

michele-riva:
@fkraushofer: renaming of `Rparams.paths.source` to `Rparams.paths.tensorleed` is done. For now, I haven't touched the `.source_dir` attribute of `DeltaCompileTask`/`RefcalcCompileTask` objects. Perhaps we can tackle that in #43 in a unified manner.

From my perspective, this PR is good to go. Would you mind reviewing?

Notice that the changes related to `Rparams.paths.work` will be done in #299.

src/viperleed/calc/sections/initialization.py
```diff
@@ -485,8 +485,8 @@ def init_domains(rp):
                 dp.sl = poscar.read()
                 dp.rp = parameters.read()                                       # NB: if we are running from stored Tensors, then these parameters will be stored versions, not current PARAMETERS from Domain directory
                 warn_if_slab_has_atoms_in_multiple_c_cells(dp.sl, dp.rp, name)
-                dp.rp.workdir = home
-                dp.rp.source_dir = rp.source_dir
+                dp.rp.paths.work = home
```
fkraushofer, line 488: In the context of rp.paths.home, I find it a bin confusing to call the work directory "home" here. Should be simply rename it locally?

fkraushofer (approved):
I would consider changing the one local variable name ('home'), otherwise it looks good to go

src/viperleed/calc/sections/initialization.py
michele-riva: Agreed. I will change that

PR #299, opened 2025-01-31
Remove Rparams.paths.work

michele-riva:
As discussed in #298. This was originally intended only as a way to fetch the tensor-LEED source files, copied in work via `job.sh`. For this reason, we would store the **main** work directory also in the `Rparams` of single structural domains in a domain calculations. We now have a cleaner mechanics for that. Using `rp.paths.work` can thus easily lead to bugs, as it may execute calculations of subdomains in the wrong folder.


michele-riva:
@fkraushofer: I've just created this PR to make it easier to look through the relevant bits of code.

src/viperleed/calc/classes/rparams/rparams.py
```diff
@@ -88,7 +88,7 @@ class _RunPaths:
     work: Path = None                                                           # TODO: see if we can get rid of this
 
     @property
-    def compile_logs(self):
+    def compile_logs(self):                                                     # TODO: Should compile_logs go to domain subfolders?
```
fkraushofer, line 91: probably yes - otherwise there may be issues with multiple files claiming the same name

fkraushofer (commented):
Concerning the last remaining "problematic" path: a try-except block around here might do the trick:
https://github.com/viperleed/viperleed/blob/f4c55956a018473002a9145fd98a7a368a38f144/src/viperleed/calc/sections/fd_optimization.py#L296-L302
 Or, otherwise, within `get_fd_r`

src/viperleed/calc/sections/refcalc.py
```diff
@@ -421,6 +421,11 @@ def refcalc(sl, rp, subdomain=False, parent_dir=Path()):
             logger.error("Exception during writePARAM: ",
                          exc_info=rp.is_debug_mode)
             raise
+        # TODO: see if we can remove paths.work here. Was introduced in         # TODO
+        # 7c6018c455d95440ae2f0b5b5c18ad1379600d85 due to a problem of
+        # not-reverting-back-to-home in case of failures. The better
+        # solution there seems to be use a try...finally block, or the
+        # execute_in_dir context manager.
```
fkraushofer, line 428: See my comment, executing in a try...finally block or context manager sounds good to me. However, I now had a closer look at 7c6018c455d95440ae2f0b5b5c18ad1379600d85 and actually, I don't see why is would make any difference if we user `rp.paths.work` here, or just an absolute global path of whatever `rp.paths.work` would have been at the time of collecting the tasks (probably just `.`). The RefcalcCompileTask and RefcalcRunTask will switch back to whatever path is passed here, there's no actual reason for that to be stored in `rp` as far as I can see.

michele-riva:
@fkraushofer: I've been working a bit on the `refcalc` and `deltas` where `rp.paths.work` was still in use. I noticed that we make a distinction between `Path.cwd()` and `comptask.basedir`/`runtask.basedir` in `run_*`/`compile_*`. However, we always do (and I think this is right) set these to be `Path.cwd()` (or rather `rp.paths.work` in some places, but the two should be the same).

So my question is: is there any reason to have `.basedir` as a public attribute? Shouldn't we just keep a private attribute to which we give the `Path.cwd()` value upon creation of the instances? Feels like the latter would simplify things a little.

fkraushofer:
> @fkraushofer: I've been working a bit on the `refcalc` and `deltas` where `rp.paths.work` was still in use. I noticed that we make a distinction between `Path.cwd()` and `comptask.basedir`/`runtask.basedir` in `run_*`/`compile_*`. However, we always do (and I think this is right) set these to be `Path.cwd()` (or rather `rp.paths.work` in some places, but the two should be the same).
> 
> So my question is: is there any reason to have `.basedir` as a public attribute? Shouldn't we just keep a private attribute to which we give the `Path.cwd()` value upon creation of the instances? Feels like the latter would simplify things a little.

Yup, agreed - I don't think it's necessary to have these attributes public, or even mutable; they just need to be assigned the original path upon creation so that they always "find their way back" in case something goes wrong.

michele-riva:
@fkraushofer: one last unused attribute: `DomainParameters.homedir` (always set to the main work directory) appears to never be used. Should I scrap that one too?

fkraushofer:
Yes, go for it! I'm not even sure what that one was originally for... I remember going back and forth with how to handle the directory structure for domains, so probably I just made it obsolete already early on without realizing it.

michele-riva:
@fkraushofer: I am now done with this one. I took the liberty of cleaning up a bit of the code in `refcalc` and `deltas`, but wrote new tests to ensure the behavior is unchanged. It would be great if you could find time to review this once again.

src/viperleed/calc/sections/deltas.py
```diff
@@ -66,189 +83,212 @@ def copy_source_files_to_local(self):
...
         logger.warning(f'Error deleting folder {workfolder.name}')
     return ''
 
 
-def compileDelta(comptask):
-    """Function meant to be executed by parallelized workers. Executes a
-    DeltaCompileTask."""
-    home = Path.cwd()
-    workfolder = (Path(comptask.basedir) / comptask.foldername).resolve()
+# TODO: see not in refcalc.compile_refcalc
```
fkraushofer, line 224: "note"?

fkraushofer (approved):
Looks good

src/viperleed/calc/sections/deltas.py
michele-riva: Oops. Fixed in 94b22dc224b84e33a762227d4df32fdb57590588

michele-riva:
> Looks good

Thanks! I'll merge it in then.

Issue #205, opened 2024-04-15
Re-evaluate bookkeeper behavior for domain calculations


amimre:
I'm copying here a [comment and discussion](https://github.com/viperleed/viperleed/pull/198#issuecomment-2048922540) by @michele-riva concerning the bookkeeper behavior in domain calculations from issue #198:

> Another important question (added above as point nr.8). The current proposed behavior of the automatic `bookkeeper` running 'around' `calc` only applies to the `cwd`. This may give some unexpected results when running a `DOMAINS` calculation. Only the 'main' directory will be automatically processed. This means that the source directories in which the domain-specific input files are located would maintain the same structure they have at the end of `run_calc` ('untouched' inputs in `root`, outputs in `OUT`). This also means that running a `DOMAINS` calculation without explicitly executing `bookkeeper` manually in each of the domains directories will effectively **not** give the same 'automatic continuation' behavior as when a single-domain calculation executes.
> 
> Two options:
> 1. point out to the user that they should manually run `bookkeeper --archive` in all the `DOMAINS` folders (plus, run again `bookkeeper --clean` to avoid problems with duplicate names for `*_ori` files).
>     **PROS**:
>       - simple to code. Just needs to look up the main `PARAMETERS` file for an uncommented `DOMAINS`.
>       - less risk of hiding some processing of a non-main folder from the user.
> 
>    **CONS**:
>       - requires more manual user input.
>       - users are likely to forget this, as they're used to the 'automatically continue' behavior of a single-domain calculation.
>       - does not allow us to add special information to `history.info`. Personally, I feel it would be good to have a tag in the `history.info` of a subdomain that points out that a certain run was executed using the structure as a `# DOMAIN of <reference to main>`. I think it is otherwise a bit complex, when looking through the `history.info`, to reconstruct that a specific entry in `history` did not correspond to a 'normal' calculation. It would also be useful for #180.
[...]

PR #302, opened 2025-02-10
Deprecate `_OUT` suffix for "state" `OUT` files, handle generated files

michele-riva:
This PR comes from a discussion between myself and @fkraushofer. It sparked from a https://github.com/viperleed/viperleed/pull/198#issuecomment-2474020727 (point 2). There, we realized that the functional change of saving the processed `POSCAR` file generated during initialization with an `_OUT` suffix is problematic: we would end up saving the incorrect `POSCAR` into Tensors files (see point 5 of the same https://github.com/viperleed/viperleed/pull/198#issuecomment-2474020727).

In an in-person discussion we then reached the conclusion that the `_OUT` suffix is unnecessary. In fact, the files are already stored in the `OUT` directory, so it should be clear to the user what is indeed an output of the calculation. Always having files in the `work` directory without the `_OUT` suffix, always guarantees that they are the newest possible files, allowing us to skip checks concerning which files should be moved where.

There are three disadvantages to not having an `_OUT` suffix:
1. An auto-generated `VIBROCC` in an un-halted `1-3` first run would be overwritten by the `VIBROCC` that comes from the optimization. The solution is to label the generated `VIBROCC` as `_generated`, and save it to `SUPP` instead, as this is a rather uncommon use case.
2. state files in `OUT` from a run like `1-3 1-3` would correspond to the **last** optimization step. This is already a "problem" since we have decided to drop the timestamps from the `_OUT` files (#198). It seems sensible, however, to only collect in the main `OUT`  the last results. Intermediate ones are collected anyway in the `workhistory` subfolders.
3. Side-by-side comparison of the original and optimized `POSCAR` files in a text editor or in VESTA does not allow a straightforward identification of which one was the input and which one the output just by looking at the file names. Users will need to look at the file paths. This should not be a major concern for a text editor. It is a bit more of an issue in VESTA, as the full file path is not visible. We deem this to be something acceptable, in favor of reducing the complexity of the suffixes.

This PR implements the necessary changes by:
- Dropping the `_OUT` suffix altogether when saving edited/generated files.
- Storing a set of generated/edited files in `Rparams`. They are moved from `work` to `OUT` at `cleanup` time.
- Storing a `VIBROCC_generated` file in `SUPP` (by duplicating the `VIBROCC` in root as soon as it is generated). It may be stored twice, once in `SUPP` once in `OUT` in an `initialization`-only run that creates it.
- Maintaining backwards compatibility in `cleanup` and `bookkeeper` in case users execute `calc` or `bookkeeper` on an old-style directory.


michele-riva:
@fkraushofer: One thing we haven't discussed: how about `R_OUT_...` files? Should we drop the `_OUT` there too?

michele-riva:
> @fkraushofer: One thing we haven't discussed: how about `R_OUT_...` files? Should we drop the `_OUT` there too? If yes, the new name would be of the form `R_<section>_R=<value>`. Does that seem reasonable?

@fkraushofer: I would be pretty much done with this, and would be ready for you to review. However, I'd like to take a decision about the above. After having given it a little thought, perhaps we can leave the `R_OUT_...` file names unchanged (mostly for the TensErLEED aficionados). What do you think? `R_OUT_<section>_R=...` or `R_<section>_R=...`?

michele-riva:
@fkraushofer: In the meanwhile, I have requested you a review.

FYI:
- I took the liberty of refactoring quite a bit of code in `cleanup`, and wrote tests to ensure the same behavior is maintained. This was necessary to support the new mechanics for "generated" files.
- Changes on documentation pages: the relevant ones are from 9baa7628fd975bc2e3cff5df47fba0fa8dc39b0d.

src/viperleed/calc/sections/cleanup.py
```diff
@@ -624,3 +589,143 @@ def preserve_original_inputs(rpars):
...
+
+def _write_manifest_file(manifest_contents):
+    """Write items in manifest_contents to file 'manifest'."""
+    manifest_contents = set(manifest_contents)
+    manifest = Path('manifest')
+    try:
+        manifest.write_text('\n'.join(manifest_contents) + '\n',
+                            encoding='utf-8')
+    except OSError:
+        logger.error(f'Failed to write {manifest} file.')
```
fkraushofer, line 729: I wonder if this still technically needs to be an error. The manifest file was important when the bookkeeper was a seperate program, as it was basically used as a to-do-list for what to copy out. Now, it is pretty much obsolete - it may still be nice to have in case something goes wrong in cleanup, or just to check what files were produced, but it's not a huge deal if it's not actually written. Maybe `logger.warning` instead?

fkraushofer (approved):
I only had a superficial look at the tests, but properly checked everything else, and it looks good to me.

I also agree with keeping the `R_OUT` files as-is.

src/viperleed/calc/sections/cleanup.py
fkraushofer: Fair point - I guess I was confused in my comment about what is what. You're right, the bookkeeper has nothing to do with it. What I was talking about is the former "job script", which used to be responsible for copying to work and back (when viperleed basically only consisted of `run_calc`), but this has now been included as a part of `calc`. In principle, as long as `run_calc` is generally called via `calc`, you would then not strictly need a `manifest` file any more.

However, you're right that these are logically separate steps, and in principle it is possible for users to construct their own implementation using only `run_calc` and relying on the manifest. So, let's leave it as is. I'll resolve the conversation then.

michele-riva:
Of course, I think found a bug in the only part of cleanup that I did not refactor/test: `move_oldruns` (looks like it probably was always there): https://github.com/viperleed/viperleed/blob/c17ed58cf7e020ecce244c50d268a58bc5ac5e20/src/viperleed/calc/sections/cleanup.py#L455-L461

Looks like the `dp`-related part misses `path=dp.workdir`. The version on master, would "organize" the wrong folder (i.e., the root, also for each subdomain, instead of the "work" of each subdomain). The current version raises `TypeError`, since I made `path` mandatory in 48fbbebcfb88a849edfa639fefab889e75c21a6c.

@fkraushofer: If you agree with the analysis above, I'd fix this. I can make a dedicated PR to refactor also the only part of `cleanup` that is untested.

fkraushofer:
Yes, agreed, this looks wrong

michele-riva:
Now that I think about it, shouldn't we rather call `move_oldruns` in all domain subfolders (if not `prerun`)? Otherwise would only create `SUPP`/`OUT` in there, but never intermediate `workhistory` folders, and we may loose the intermediate results from the domains.

fkraushofer:
> Now that I think about it, shouldn't we rather call `move_oldruns` in all domain subfolders (if not `prerun`)? Otherwise we would only create `SUPP`/`OUT` in there, but never intermediate `workhistory` folders, and we may loose the intermediate results from the domains.

That would probably be the clean way to handle it, yeah... it is somehow linked to #205. So far, the implementation for domains just follows a "keep it simple" philosophy, where hardly anything is handled automatically, and the user is expected to just avoid any fancy "queued" jobs if they need all the details. I guess it wouldn't be a huge effort to make both the cleanup and the bookkeeper run as normal in each of the domain subfolders - but there's likely a bunch of edge cases (e.g. files that don't exist in the subfolders) that then need to be considered.

michele-riva:
Perfect. If you have time, I'd add this to our TODO list for today's call.

PR #304, opened 2025-02-18
Finish refactoring cleanup

michele-riva:
- Split up `move_oldruns` into simpler helper functions
- Improve docstrings
- Longer names everywhere
- Improve readability of the implementation of helper methods, without functional changes (exceptions: see below)
- Use `Path` everywhere instead of `os`
- Catch only sensible exceptions to prevent hiding bugs

**Functional changes:**
- fix: (temporary) call `organize_workdir` in all domain subfolders. For the final solution that will be adopted, see https://github.com/viperleed/viperleed/pull/302#issuecomment-2663716503
- Concerning files to store in `workhistory` during `prerun`: (i) skip `rpars.files_to_out` in addition to `manifest`; (ii) return SUPP/OUT only if they exist, avoiding annoying `logging.WARNING` for, e.g., init-only runs.
- Small change to `workhistory` subfolder names in case no segment with an abbreviation was run: avoid double underscore (was `tTTT.rRRR__TIMESTAMP`).
- Support workhistory folders with more than 3-digit numbers

All of the above is now tested. Module coverage is now 100%.

### DIFF
--- a/tests/calc/sections/cleanup/test_organize_workdir.py
+++ b/tests/calc/sections/cleanup/test_organize_workdir.py
@@ -445,7 +445,7 @@ class TestZipSubfolders:
                     archive=True,
                     compression_level=2)
         expect = self.packed_all
-        # The non-matching folder should sty where it is
+        # The non-matching folder should stay where it is
         expect[self.folder].update(tree[self.folder])
         assert clean == expect
         assert not caplog.text
--- a/tests/calc/sections/cleanup/test_organize_workdir.py
+++ b/tests/calc/sections/cleanup/test_organize_workdir.py
@@ -11,7 +11,6 @@ import logging
 import shutil
 from zipfile import ZipFile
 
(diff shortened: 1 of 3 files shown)


michele-riva:
@fkraushofer: This one does not yet contain the propagate-to-domain features we discussed earlier. I will build on this one to implement that in a separate PR.

tests/calc/sections/cleanup/test_organize_workdir.py
```diff
@@ -451,6 +445,7 @@ def test_invalid_name_format(self, run, workdir, caplog):
                     archive=True,
                     compression_level=2)
         expect = self.packed_all
+        # The non-matching folder should sty where it is
```
fkraushofer, line 448: typo: stay

fkraushofer (approved):
Good to go, found only a typo

PR #311, opened 2025-03-06
Prevent bookkeeper logs to spill over to calc

michele-riva:
In the implementation of `calc`+`bookkeeper` before this PR, log messages emitted by the latter would end up in the main `calc` log file if `run_calc` exits before calling `section_loop`. That's because we did not close handlers associated with the `calc` logger under these conditions. Since `bookkeeper` is a sub-package of `calc`, its logger is a "child" of the `calc` one.

This PR does:
- refactor `calc.run` into helper functions to improve readability
- add comprehensive tests for 100% coverage of the module
- close handlers of the calc logger also upon early exit in case of either error or a halting condition.

### DIFF
--- a/src/viperleed/calc/bookkeeper/cli.py
+++ b/src/viperleed/calc/bookkeeper/cli.py
@@ -119,7 +119,7 @@ class BookkeeperCLI(ViPErLEEDCLI, cli_name='bookkeeper'):
     def __call__(self, args=None):
         """Call the bookkeeper with command-line args."""
         parsed_args = self.parse_cli_args(args)
-        bookkeeper = Bookkeeper(cwd=Path.cwd().resolve())
+        bookkeeper = Bookkeeper()
         mode = getattr(parsed_args, 'mode', BookkeeperMode.ARCHIVE)
         kwargs = {
             'requires_user_confirmation': not parsed_args.skip_confirmation,
--- a/src/viperleed/calc/files/phaseshifts.py
+++ b/src/viperleed/calc/files/phaseshifts.py
@@ -485,7 +485,7 @@ def __check_consistency_element_order(rp, sl, phaseshifts,
     return may_have_wrong_phaseshifts
 
 
-def writePHASESHIFTS(firstline, phaseshifts, file_path=Path()/'PHASESHIFTS'):
+def writePHASESHIFTS(firstline, phaseshifts, file_path='PHASESHIFTS'):
     """Takes phaseshift data and writes it to a PHASESHIFTS file."""
     _file_path = Path(file_path)
     output = firstline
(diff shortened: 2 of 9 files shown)


michele-riva:
I have a couple more questions that came up while looking at the code:
1. @fkraushofer: I believe calling `close_all_handlers` in `cleanup` has no effect: https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/sections/cleanup.py#L491-L494 In fact the logger in the `cleanup` module is not the main `calc` logger, and has no associated `handlers`. This part seems to have been there all along. It made sense when the whole code was in the `tleedm.py` module, as then the `logger` was the correct one. I would suggest removing it for clarity. Does this make sense?
2. @fkraushofer, @amimre: I see multiple calls to `logging.shutdown` throughout the code: https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/utilities/poscar/attach_bulk.py#L170-L173 https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/run.py#L205-L208 https://github.com/viperleed/viperleed/blob/d43f96a9f85f1f60de39e875a195574fc9ddf17e/src/viperleed/calc/sections/cleanup.py#L491-L494 However, the [docs of the logging module](https://docs.python.org/3/library/logging.html#logging.shutdown) state that `logging.shutdown` (i) should be called only at **application exit**, and (ii) is automatically registered `atexit` whenever `logging` is imported, so it does not really need to be called. I would suggest we remove those calls. However, I see a comment by @amimre in `calc.run.run_calc` about "breaking unit testing". Could you elaborate on what was the problem? The commit that introduced this is 4fe4f50800b324648cdef744e4237d9e281fe4aa.

michele-riva:
> [...] However, I see a comment by @amimre in `calc.run.run_calc` about "breaking unit testing". Could you elaborate on what was the problem? The commit that introduced this is [4fe4f50](https://github.com/viperleed/viperleed/commit/4fe4f50800b324648cdef744e4237d9e281fe4aa).

Concerning this: I ran tests with `shutdown` commented out, and I don't seem to see any problem.

fkraushofer:
As far as I can tell, all the duplicate `close_all_handlers` / `logging.shutdown()` are remnants of `tleedm` being organised in various different scripts. I agree, all of that should be consolidated, and everything that may happen in the middle of execution should be removed.

amimre:
I believe part of this came from separating the bookkeeper from the main part, in order to allow the bookkeeper to have a separate log.
The "breaking unit tests" was related to usage of the `caplog` fixture in pytest, I think, i.e. tests that check for the correct log outputs.

amimre:
Tests look good on my side.

amimre (approved):

michele-riva:
@fkraushofer: since @amimre had no problems with the test, I'd go ahead and merge this. Let me know if you have any comments (or are planning to take a look at it). Happy to include them later.

Issue #131, opened 2023-11-28
Utility for calculating R-factors between two CSVs


fkraushofer:
It would be nice to have a viperleed utility that can quickly calculate the R-factor (overall and for individual beams) between two datasets (read from CSV). This would require input of a V0i and a V0r shift range (although one could supply sensible defaults, or read from PARAMETERS if present), and possibly some form of beam equivalence. For beam equivalence, the default could just be that only beams with the same names are equivalent. My usecase is essentially fingerprinting, so comparing multiple measurements of the same sample after different treatments. I have no problem putting a simple PARAMETERS file as input, but would rather not supply a POSCAR, VIBROCC etc.

Something very similar is already implemented by @schmid-iap in the ImageJ plugins, called "R factor between data sets". This does exactly what I want, beam equivalence is done via a pattern file (also a sensible solution for the utility, I would say). The nice thing about implementing the same as a python utility would be that plots can be generated in the same style as for the calculated beams, with R-factors printed into the plots for each of the beams. I would say the PLOT_IV parameters could either be input via a GUI, or entered interactively in the shell, or be read from a PARAMETERS file if present (which would only have to contain these parameters, but could also be accepted if it contains more, e.g. from a previous calculation). This would make the utility somewhat related to https://github.com/viperleed/viperleed/issues/87, though I'm not sure whether it's exactly the same (would we always want R-factor calculation there? I suppose we do want them printed in most cases... but for computed results, we could also read from previously calculated output).

As already briefly discussed by e-mail with @michele-riva and @amimre, there currently seems to be some redundancy between the rfactor_from_csv function in viperleed_from_ase.py (which does most of what I would want) and the run_new_rfactor function in tleedmlib.sections.rfactor; the latter currently requires both a Slab object (for determining beam equivalence) and an Rparams object (for all the input, but also for storing some results). I would suggest moving the R-factor-related functions in viperleed_from_ase to tleedmlib.leedbase, and to refactor such that the run_new_rfactor uses those functions for the computation, with some Slab/Rparams handling before and after. The utility could then run completely based on these functions in tleedmlib.leedbase.

PR #319, opened 2025-04-03
Automatically remove work directory upon successful execution

### LINKED ISSUE #316: Automatically `--delete-workdir`
The current default for the `--delete-workdir` CLI argument of `calc` is `False`. It probably would be nicer to change the default behavior to automatically delete `work` unless `calc` exited with an error condition.

I'd like to hear @fkraushofer's and @amimre's opinion on this.

michele-riva:
Fixes #316.

The implementation is slightly different than I originally suggested in https://github.com/viperleed/viperleed/issues/316#issuecomment-2772040216:
- if the `--keep-workdir` flag is not given, the work directory is deleted unless `run_calc` exited with any error condition.
- if the `--keep-workdir` flag is given, never delete the work directory, irrespective of the exit code.

The reason not to implement it as I originally suggested is that we would then need to make the `--keep-workdir` an argument rather than a flag, so people would need to call with one of the following:

```
viperleed calc                       # work deleted if no errors
viperleed calc --keep-workdir True   # always keep work
viperleed calc --keep-workdir False  # always delete work
```

Concerning the "error condition": the current implementation will consider a `KeyboardInterrupt` during `section_loop` as an error condition, so the work directory will stay. I'd like your thoughts about this.

### DIFF
--- a/doc/content/calc/example_systems/Ag(100).rst
+++ b/doc/content/calc/example_systems/Ag(100).rst
@@ -195,8 +195,6 @@ Your root folder should now look like this:
     │   ├── original_inputs/
     │   │   └── ...
     │   └── ...
-    ├── work/
-    │   └── ...
     ├── EXPBEAMS.csv
     ├── history.info
     ├── IVBEAMS
@@ -210,10 +208,7 @@ Your root folder should now look like this:
     └── viperleed-calc-<timestamp>.log
 
 Folders :file:`OUT` and :file:`SUPP` always contain the main
-:ref:`results <files>` of the calculation. Folder :file:`work`
-is where |calc| actually runs. It is kept for debugging purposes,
-and you can safely delete it (use the ``--delete-workdir``
-:ref:`command-line argument <cli_calc>` to do this automatically).
+:ref:`results <files>` of the calculation.
 Files :ref:`ivbeams`, :ref:`phaseshifts`, and :ref:`vibrocc` were
 automatically created by |calc|. :file:`POSCAR_user` is an exact
 copy of the :ref:`poscar` file given as input the first time |calc|
--- a/doc/content/calc/how_to_run.rst
+++ b/doc/content/calc/how_to_run.rst
@@ -109,9 +109,11 @@ A typical call may look like this:
 
             viperleed calc -w %work_path% -t %tensorleed_path%
 
-The directory at ``work_path`` is the one where the calculation will be
+The directory at :file:`work_path` is the one where the calculation will be
 executed and all temporary files will be stored. It is created if it does not
-exist. The ``tensorleed_path`` is the path to the :ref:`install_tensorleed`.
+exist, and automatically deleted upon successful termination (unless the ``-k``
+command-line flag is specified).
+The ``tensorleed_path`` is the path to the :ref:`install_tensorleed`.
 If the ``-t`` option is not given, ViPErLEED looks for the TensErLEED source
 code under the path specified by the :envvar:`VIPERLEED_TENSORLEED`
 :term:`environment variable` (see also :ref:`set_envvar`).
@@ -164,7 +166,9 @@ a :ref:`manifest` file that lists the files and directories to be copied back.
 :numref:`list_organization_output` shows an example of the directory tree
 after a run. The :file:`my_work` directory may be on a different file-system
 path (including a different drive, a network path, or a remote server) or be
-a subfolder of :file:`my_surface`.
+a subfolder of :file:`my_surface`. It is shown mainly for illustration
+purposes, as the work directory is automatically deleted after a successful
+execution (unless the ``-k`` command-line flag is given).
 
 .. _list_organization_output:
 .. code-block:: console
@@ -207,7 +211,7 @@ a subfolder of :file:`my_surface`.
     ├── VIBROCC              <-- output, created or edited by calc
     └── viperleed-calc-<timestamp>.log
 
-    my_work/
+    my_work/                 <-- deleted upon successful termination
     ├── manifest
     └── ...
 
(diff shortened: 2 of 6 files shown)


amimre:
Sounds good!
I'd keep the Keyboard interrupt as an error as it constitutes a non-normal exit condition and may be useful for quick debugging.

I'd suggest maybe adding a shorthand flag (e.g. `viperleed calc -k`).

michele-riva:
After implementing this, I now have some reservations as to whether this was actually a goof idea. I'd like an opinion from @fkraushofer and @amimre about this:

One reason against automatically deleting the `work` directory by default is that this causes an extra "overload" for copying and unpacking `Tensors` and `Deltas`. I recall that @LutzHammer made a point a while back that this was quite a bottleneck for the "quick" calculations (e.g., running an _R_ factor). Not sure what's your feeling about this. I'm also not sure of whether it would be worth to automatically remove everything from the `work` directory **except** for the `Tensors` and `Deltas` folders.

fkraushofer:
I think that bottleneck is fairly wide. If you're actually just calculating R-factors, yes, the unpacking is a relevant time factor (though I'm not sure if we even do unpack in that case - maybe to verify that the state is unchanged..? - anyway, certainly no Deltas). If that's actually a concern, and users really can't wait 30 seconds, then we could offer a faster R-factor calculation with something like #131. I can't really think of any other application where unpacking is a relevant contributor to execution time.

The bigger issue would be if we always copy the __entire__ Tensors and Deltas directories, including all the old versions. This was certainly the case way-back-when (and my main reason for usually keeping `work`), because copying stuff to work was done in the job script, which didn't know yet which Tensors/Deltas would be used in the calculation. The new CLI implementation would make it possible to have smarter approaches (e.g. don't copy anything to work at the beginning, only unpack directly into `work` once needed). Not sure if this is already the case.

Personally, I'm pretty sure I'll always run in `-k` mode anyways - but I also always have debug mode on, which is also not the default. I think as long as such a non-default option exists, people who want the absolute best performance and most information can anyway use that - but it's still probably nicer to hide all this extra stuff away by default to not overwhelm people.

amimre:
I agree with Florian and think we should have the default of deleting.
As far as I remember the complaint for Lutz was because he was dealing with the huge tensors for the (10x10) system. For more "normal" sized structures, decompression should be pretty fast.

Also, for systems where this is really a concern, we have ZIP_COMPRESSION_LEVEL which can speed it up.

For normal use it's probably *more* annoying to have the work directory with Tensors and Deltas that take up extra disk space if not removed manually.

michele-riva:
Alright, thanks guys for the feedback. Then let's go ahead with this. I jot down some notes below.

> I think that bottleneck is fairly wide. If you're actually just calculating R-factors, yes, the unpacking is a relevant time factor (though I'm not sure if we even do unpack in that case - maybe to verify that the state is unchanged..? - anyway, certainly no Deltas).

We do in fact unpack missing `Tensors` as we need them for the _R_. If we had removed `work` at the end of the "previous" run, then there is no `Tensors` folder in `work`. https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/sections/rfactor.py#L88-L93

This part is skipped if we already find `fd.out`/`spec.out` in work.

> If that's actually a concern, and users really can't wait 30 seconds, then we could offer a faster R-factor calculation with something like #131. I can't really think of any other application where unpacking is a relevant contributor to execution time.

Perhaps running an explicit `superpos`? There, we would need to fetch `Deltas`. On a second look, though, I believe we do not really support this, as we rely on the presence of an `SD.TL` file in `work`. https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/sections/superpos.py#L225-L237

> 
> The bigger issue would be if we always copy the **entire** Tensors and Deltas directories, including all the old versions. This was certainly the case way-back-when (and my main reason for usually keeping `work`), because copying stuff to work was done in the job script, which didn't know yet which Tensors/Deltas would be used in the calculation. The new CLI implementation would make it possible to have smarter approaches (e.g. don't copy anything to work at the beginning, only unpack directly into `work` once needed). Not sure if this is already the case.

The default for the `--all-tensors` CLI argument is `False`: https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/cli.py#L122-L128

so we only pull the most recent `Tensors` file into work: https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/cli.py#L168-L189

Notice that we don't really unpack the `Tensors` at the time the CLI is called. Unpacking happens in the relevant sections: https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/sections/deltas.py#L366-L370 https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/classes/rparams/domain_params.py#L103-L114 https://github.com/viperleed/viperleed/blob/ee723272cc09ab6126f5dd3fe6ace051dbaa87ab/src/viperleed/calc/sections/rfactor.py#L88-L93
 
> Personally, I'm pretty sure I'll always run in `-k` mode anyways - but I also always have debug mode on, which is also not the default.

This made me think about a new feature that we may support: a global configuration file for "defaults" that allows people to avoid specifying over and over the same thing. I believe @LutzHammer has asked for something along these lines for `PARAMETERS` defaults. I think we can easily extend it to be more general. I have opened #320 with some suggestions.

> I think as long as such a non-default option exists, people who want the absolute best performance and most information can anyway use that - but it's still probably nicer to hide all this extra stuff away by default to not overwhelm people.

Fine by me.

fkraushofer (dismissed):

michele-riva:
@fkraushofer: would you mind taking a look at the new changes on this one too? Then I can merge it as well into 0.13.0.

fkraushofer (approved):

michele-riva:
Sorry, @fkraushofer, once more...

fkraushofer (approved):

Issue #215, opened 2024-06-30
Bookkeeper does not behave correctly when TENSOR_INDEX is used


michele-riva:
Reported by @fkraushofer:

> Setting TENSOR_INDEX to something older (basically I want to load and older configuration and avoid doing a new refcalc), history.info does not reflect the actual Tensors used.
> ```
> # TENSORS   21
> # JOB ID    2
> # RUN       0 11
> # TIME      04.06.24 16:02:07
> # R REF     0.1537
> # FOLDER    t021.r002_240604-160207
> Notes: plotting only, full energy range.
> ```
> but actually TENSOR_INDEX = 18 in this run.

A short conversation revealed that the behavior was likely never correct for a specific `TENSOR_INDEX` value. This is not yet fixed in the overhaul of bookkeeper in `bookkeeper-dev` that will enter v1.0.0.

PR #325, opened 2025-04-08
Propagate bookkeeper to domains

### LINKED ISSUE #205: Re-evaluate bookkeeper behavior for domain calculations

michele-riva:
Fixes #205.

### Main changes
- [x] `Bookkeeper` accepts a list of additional paths in which to execute.
- [x] calc `cli` passes those paths in after a `DOMAIN` calculation is finished. They are the paths to the subfolders of the directory in which `calc` ran that contain the output of each domain. The paths are inferred from the `manifest` file.
- [x] `Bookkeeper` adds domain information to metadata files (using the "main" history folder, i.e., not those coming from `workhistory`; only the main folder of each `history` contains domain information):
  - **main directory**: which domains were also calculated, as a list of `(relative_path, hash_of_history_folder)` for each domain.
  - **each subdomain**: information about the run in the main directory, as a `(absolute_path_to_main, hash_of_history_folder_of_main_run)` pair
- [x] If not passed any paths, `Bookkeeper` should check the metadata files in `history` to find out whether the current one was a domain calculation, and run in the same mode also in the domain subfolders. Only the most recent `history` folder is checked. This allows the first `bookkeeper --clear` call in `calc` to correctly handle the domain paths too. In fact, at that point we can't rely on the existence of a `manifest` file, as that one is not copied back from `work`, and the latter may not exist any longer by the time execution starts again.
- [x] If `Bookkeper` is run manually in a domain subfolder (say, `Domain_1`), warn about running manually also in the main one as well as siblings. NB: Domain information will not be retained for the `history` folder created this way. This should not pose a problem, as the normal workflow would be `calc` (runs `--clear` before, `--archive` after) + manual `bookkeeper` (in `Domain_1` only, e.g., in `--discard` mode) + `calc`. The domain info is not retained in the manual invocation only if the `--archive` run after the first `calc` fails. Even if that's the case, the `Domain_1` folder is cleared already. The `--clear` call before the second `calc` will deal with the main and the other siblings (potentially adding incomplete domain info, i.e., missing `Domain_1`). However, the `--archive` at the end of the second `calc` uses explicit paths, and would restore the correct domain information.
- [x] If `Bookkeeper` is run manually in a domain subfolder, use information from the main log file, unless a log file exists in that folder (this means `calc` was explicitly called in there). This prevents double-storage of history contents and correctly processes `workhistory` folders (see https://github.com/viperleed/viperleed/pull/325#issuecomment-2837979609 for more details).
- [x] When running in `--discard-full` mode, check preemptively whether it can be executed on **both the main folder as well as all domains** before actually proceeding. This is because `--discard-full` is destructive.
- [x] If domain subfolders are not found, warn but still proceed. Once "Future developments" point 2 is done, this will fail instead, and require an explicit `bookkeeper --fix` call. See #336.
- [x] If the main path of a domains calculation does not match the one stored in the `metadata` files, ask user confirmation. Once "Future developments" point 1 is done, this will fail instead, and require an explicit `bookkeeper --fix` call. See #336.
- [x] If a multi-domain run fails, warn users about calling `bookkeeper` explicitly in all subfolders.

### Minor changes
- [x] Do not to complain about "missing" `STATE_FILES` when they're not supposed to be there in the main folder of a `DOMAIN` calculation.

### Points to be discussed
1. **REJECTED**: `Bookkeper` may also auto-execute if it was run in a domain subfolder: it also runs on all other domains as well as the main one, in the same mode. This is to avoid, e.g., `-df` on one domain and not the others. Not sure if this should be restricted to `-df` mode only. This, however, poses some user restrictions: one would not be able to explicitly limit a `bookkeeper` run to only a specific domain subfolder. Maybe a better alternative would be to warn about the risks and ask for user confirmation to proceed.
2. **POSTPONED**: Add a `--subdomains`/`-s` (default: `True`) CLI argument. Pending user feedback on whether it feels necessary to **not** automatically run in domain subfolders when invoked from the main one (and not complain when invoked in one of the subfolders). See #335 for progress.

### Future developments
Probably to be tackled in separate PRs.
1. Allow `--fix` to update the paths of the main directory in `metadata` files to the one in which `bookkeeper` is called. #336
2. `--fix` also takes care of the case in which domain subfolders have been renamed relative to those stored in the metadata files of the `history` of the main directory. #336


amimre:
For the point above:
Does the bookkeeper detect reliably if it's called in a subdomain directory?
If so, I would say if invoked in the main directory it applies to all subdomains by default. If ran in a subdirectory, it only applies to that be gives a warning (but still do it).

If it seems necessary, we can add a `--subdomains/-s` flag that defaults to True.

Checking wether -df can be exececuted on both main and sub is a good idea probably.

michele-riva:
> For the point above: Does the bookkeeper detect reliably if it's called in a subdomain directory?

Yes, as long as `bookkeeper` was called at least once in the main directory, which should now happen automatically whenever `calc` runs (via the `cli`).

michele-riva:
> Hm, I would say it should be enough to check whether the last run was a domain calculation. That kind of goes in hand with the logic of the bookkeeper commands always referring to the previous run.

That was also my feeling, but I just wanted to check your opinion.

> If the subdomain directories have been renamed and can't be found, I would warn but still perform the requested action.

Sounds good.

> Unless that breaks the logic somehow?

I can perhaps implement one extra `--fix` action that would recognize this renaming. I keep track of the hashes of the history folders, so it would be possible to find out which domain folder has been renamed to which other.

fkraushofer:
> Hm, I would say it should be enough to check whether the last run was a domain calculation. That kind of goes in hand with the logic of the bookkeeper commands always referring to the previous run.

Agreed

> For the point above: Does the bookkeeper detect reliably if it's called in a subdomain directory? If so, I would say if invoked in the main directory it applies to all subdomains by default. If ran in a subdirectory, it only applies to that be gives a warning (but still do it).
> 
> If it seems necessary, we can add a `--subdomains/-s` flag that defaults to True.
> 
> Checking wether -df can be exececuted on both main and sub is a good idea probably.

Also agreed

> I can perhaps implement one extra --fix action that would recognize this renaming. I keep track of the hashes of the history folders, so it would be possible to find out which domain folder has been renamed to which other. We can then decide whether we want to go ahead or to require people to call bookkeeper --fix, then call again in the desired mode.

Certainly a nice feature, but not suuuper necessary I feel... and it sounds like the sort of thing that opens up a whole can of worms of edge cases. But I have no real objections if you want to do it.

michele-riva:
One more implementation question for @fkraushofer and @amimre. In the current implementation, in the `metadata` files of the `history` folders of the domain subfolders I store the **absolute** path to the main directory. This may be not ideal if people rename the top-level directory, which seems a relatively innocent change. However, this may also signal that people have copied the domain subfolder somewhere else (e.g., to proceed with calculations on that one only, or to use it in a different system -- In that case, I'd assume people would also clean up the `history`, but you never know). Do you feel I should fall back to `Path('..')` in case that folder is renamed? I'm not sure whether always storing '..' (or not storing anything and always assume it's `..`) is a good idea in general for the copy-somewhere-else case mentioned above.

fkraushofer:
Hmm, I think we definitely want some backup plan for when users move their whole LEED-I(V) data tree, or some large part of it - not an everyday occurance in my experience, but it definitely happens (for example moving to a different computing cluster / from a local server to a cluster). So, falling back on `..` sounds like a good idea - or one could have some way for `--fix` to identify this path change and fix the global paths everywhere, but that seems more difficult.

About just assuming `..` or storing the name or `..`, I'm not sure - it seems a good idea to at least have some sanity check about the name of the parent directory to avoid mis-identifying some other case. Perhaps in all of these cases, output a warning and ask for user confirmation before proceeding?

amimre:
Would it be an option to store the relative path to the main directory instead?

michele-riva:
> Would it be an option to store the relative path to the main directory instead?

Unfortunately, the relative path (with respect to a domain subfolder) is always `'..'`, as, after `calc` executes, domain results are always collected as subfolders of the one in which `calc` ran.

michele-riva:
> Hmm, I think we definitely want some backup plan for when users move their whole LEED-I(V) data tree, or some large part of it - not an everyday occurance in my experience, but it definitely happens (for example moving to a different computing cluster / from a local server to a cluster). So, falling back on `..` sounds like a good idea - or one could have some way for `--fix` to identify this path change and fix the global paths everywhere, but that seems more difficult.

I like the `--fix` option idea. I think it would be easy to implement: it just needs to check consistency of the various contents of the `history` folders (to make sure that all folders that should be in place are still there), but then it can easily update all the `metadata` files. Not sure if I want to implement this now, though. I'll put it on the TODO list and see how far I get. 

> About just assuming `..` or storing the name or `..`, I'm not sure - it seems a good idea to at least have some sanity check about the name of the parent directory to avoid mis-identifying some other case. Perhaps in all of these cases, output a warning and ask for user confirmation before proceeding?

Sounds like a plan. We can later on fail and require a `--fix` run.

michele-riva (commented):
Marking a point in time for myself.

michele-riva (commented):
One more time marker for myself.

michele-riva:
@fkraushofer @amimre

Hey guys, I just ran into a corner case while writing some tests, and I'd like your opinion on how to proceed. Assume an `ARCHIVE` run fails at the end of the first execution of a `DOMAIN` calculation. If users start again `calc` (or manually invoke `bookkeeper`), only the main folder would be processed. This is because there is no information yet about domains stored in `history`. How to deal with this situation? It is probably not ideal to "just let it slide silently", as the results of the calculation where `bookkeeper` failed would be lost. I have some heuristic in place to find subfolders that "look like domains". In principle I could select those and ask for user confirmation whether these are indeed domain subfolders. Does that sound reasonable?

michele-riva:
> It sounds reasonable, but I guess it depends on how reliable these heuristics are. I could imagine it quickly becoming quite annoying if there are lots of false positives. Would it for example think that another "main" folder nested in a normal one "looks like a domain"? That seems like quite a normal thing to do (I've done this e.g. when branching off some modification into a low-precision test to see what happens)

Yes, indeed that would be the case. The heuristic is very very basic. It only checks folder names ("Domain_*" is a domain), and files/folders contained in the subfolders (any of "history", "history.info", "SUPP", "OUT", a calc log file would mark a subfolder as a domain). I'm not sure we can be fancier than this, though, for filtering out your case. Perhaps the only good way to ensure a clear identification that "this is the root of a domain calculation" is looking at the `PARAMETERS` file. I tried to steer away from this not to have to deal with all the corner cases of "which `PARAMETERS` file?" (root? `SUPP/original_inputs`? last `history` folder root? `SUPP/original_inputs` of last `history` folder? Any of the previous for any older `history` folder?), but that may be necessary.

amimre:
How about a warning in case the bookkeeper fails and know it's a domain calculation (or doesn't know yet) along the lines of "WARNING: Bookkeeper failed and was not able to process domain-directories. If you are executing a domain calculation, make sure to manually invoke the bookkeeper in the domain subdirectories"?

fkraushofer:
Hrm, yeah. I'd rather not auto-process a folder or stop every time a user has an (entirely legitimate!) directory structure where things other than domains are nested in a home folder. Perhaps just pop up a warning at the point of failure then, as @amimre suggests, and generally don't cover this case in the start-up-execution of bookkeeper

In any case, I think it should be a relatively rare event. And going for full coverage of all cases of "bookkeeper fails in end execution and user starts a new calc without first cleaning up manually" is a bit of a fool's errand I feel. Maybe another thing to put on the list for the "fix" tag, but also at some point we need to rely on the user to read error messages and act accordingly...

michele-riva:
Perfect, thanks for the feedback.

michele-riva:
@fkraushofer @amimre 

I noticed a non-ideal behavior of `bookkeeper` when it is executed manually in a domain subfolder. Since there is no calc log file in the domain subfolders, `history` folders will always be labeled as `*_moved-<date_time_at_which_bookkeeper_ran>`. This feels a bit confusing, as the "moved" was intended originally for labeling folders on which execution is duplicated. Also, this will cause problems with `workhistory` folders: since there is no log, they are not pulled into `history`.

I would like your take on how to proceed: should a subdomain always use the timestamp from the log file of the main folder? I would have to check the feasibility of this, but that seems like the best way to go.

michele-riva:
> @fkraushofer @amimre
> 
> I noticed a non-ideal behavior of `bookkeeper` when it is executed manually in a domain subfolder. Since there is no calc log file in the domain subfolders, `history` folders will always be labeled as `*_moved-<date_time_at_which_bookkeeper_ran>`. This feels a bit confusing, as the "moved" was intended originally for labeling folders on which execution is duplicated. Also, this will cause problems with `workhistory` folders: since there is no log, they are not pulled into `history`.
> 
> I would like your take on how to proceed: should a subdomain always use the timestamp from the log file of the main folder? I would have to check the feasibility of this, but that seems like the best way to go.
> 
> **Edit**: I have investigated this a bit, and it would be possible to use the root log file even if we detect that we're running in a subdomain.

Hi all. In the meanwhile I convinced myself that this is the most reasonable way and went ahead to implement this. It can be reverted if you don't agree.

fkraushofer:
Agreed, that seems the best way to handle this case!

michele-riva:
@fkraushofer @amimre 

Hi all, I noticed one non-ideality in a multi-domain calculation. Take, for example the folder structure in [Listing 27](https://www.viperleed.org/latest/content/calc/sections/domain_calculation.html#id1) of our documentation. There, the root folder of `my_domain_2` contains some input files that will not be used, in favor of those that come from `Tensors/Tensors_005.zip`. Running `calc` in this tree now processes all the domain subfolders at the end of the execution. However, since there are input files in the root of `my_domain_2`, they are the ones renamed to `_ori`. Normally this is not an issue, as those are usually the same as for the latest `Tensors` (as long as the last thing that people executed was a `refcalc`). I have now documented this, but that made me think of whether the strategy of pulling inputs from the most recent `Tensors` is really the right way to go:

Suppose someone ran a one-domain structure optimization in `my_domain_2`, without a final `refcalc`. What I would expect from an automatic "continuation" mode is for the multi-domain calculation to use the result of that optimization (as it would in a normal single-domain case), and rather identify that domain as a "`refcalc`-required" one. Perhaps we should have a way for users to decide whether they want to pull inputs from `Tensors` in case there are "bare" inputs in the root (and those differ from the ones in the most recent `Tensors`)? What do you think?

michele-riva:
@fkraushofer @amimre 

Hi, I have one last question before this is done. When running in subdomains, `OSError` may happen in some of them (e.g., user does not have permission). Currently, I stop at the first `OSError`. This may leave other domains, which could potentially be correctly processed, unprocessed. Should we keep it this way, or should we attempt processing all and re-raise at the end? In this case, should we register the domains that we processed as subdomains and leave the ones that failed out, or should we not register any? (In both cases, the information in the `metadata` of the root will be incomplete, with obvious implications for re-running in the same tree right after the failure.)

fkraushofer:
> **Edit2**: On second thought, this is even worse, I fear. If people run a domain optimization twice in the same tree, without an explicit `refcalc` at the end of the first one, the second domain optimization re-starts from the same inputs of the previous one. This last case can be auto-fixed by replacing the last `4` in `RUN` with a `1-3 1` (or `2-3 1` whatever applies) instead of a `1-3`/`2-3`, de facto forcing a final `refcalc` on all domains.

Hmm, this is a good point. I'm not sure that the solution is to force a refcalc, though. Rather, we should perhaps rethink the part where we automatically user a Tensors file if present, and rather decide based on what else is around. In your suggested approach, where we take the input files from the `Tensors.zip`, couldn't we then simpley compare this to the `POSCAR` and `VIBROCC` files (if those are there)? If the files in the domain directory differ from the ones in the `Tensors.zip`, then this means they were either manually modified or are the result of a previous optimization. In both cases, I think we should use these files, rather than the `Tensors.zip` (unless the user has specificly supplied the path to the `zip` rather than the directory).

fkraushofer:
> @fkraushofer @amimre
> 
> Hi, I have one last question before this is done. When running in subdomains, `OSError` may happen in some of them (e.g., user does not have permission). Currently, I stop at the first `OSError`. This may leave other domains, which could potentially be correctly processed, unprocessed. Should we keep it this way, or should we attempt processing all and re-raise at the end? In this case, should we register the domains that we processed as subdomains and leave the ones that failed out, or should we not register any? (In both cases, the information in the `metadata` of the root will be incomplete, with obvious implications for re-running in the same tree right after the failure.)

I feel like not processing any is the right call here - having some directories processed and others not seems more confusing, and more of a headache to fix.

michele-riva:
> > **Edit2**: On second thought, this is even worse, I fear. If people run a domain optimization twice in the same tree, without an explicit `refcalc` at the end of the first one, the second domain optimization re-starts from the same inputs of the previous one. This last case can be auto-fixed by replacing the last `4` in `RUN` with a `1-3 1` (or `2-3 1` whatever applies) instead of a `1-3`/`2-3`, de facto forcing a final `refcalc` on all domains.
> 
> Hmm, this is a good point. I'm not sure that the solution is to force a refcalc, though. Rather, we should perhaps rethink the part where we automatically user a Tensors file if present, and rather decide based on what else is around. In your suggested approach, where we take the input files from the `Tensors.zip`, couldn't we then simpley compare this to the `POSCAR` and `VIBROCC` files (if those are there)? If the files in the domain directory differ from the ones in the `Tensors.zip`, then this means they were either manually modified or are the result of a previous optimization. In both cases, I think we should use these files, rather than the `Tensors.zip` (unless the user has specificly supplied the path to the `zip` rather than the directory).

Yes, I agree. We could be even "fancy" and only take the ones in the root if they are newer than those in the `Tensors.zip`. I will open a separate issue though, as this is not `bookkeeper`-related.

michele-riva (commented):
One more time marker before I ask for a review.

michele-riva:
@fkraushofer, would you mind taking a look at this one?

src/viperleed/calc/bookkeeper/utils.py
```diff
@@ -19,6 +19,17 @@
...
 
+def ask_user_confirmation():
+    """Request user input on an action."""
+    while True:
+        reply = input('Are you sure you want to proceed (y/N)? ')
+        reply = reply.lower()
+        if not reply or reply.startswith('n'):
+            return False
+        if reply.startswith('y'):
+            return True
```
fkraushofer, line 30: This may be problematic if the bookkeeper is executed automatically by `calc`, and `calc` runs on a compute clusters managed by e.g. `slurm`. In such a case, there is no user to give an input. I think it will just get stuck in this state then. Can we have smoe override-mode attribute, which can be passed to `calc` and from there through to the bookkeeper, to automatically return `y` to all?

fkraushofer: Alternatively, if there is no response, we should time out after some time (perhaps 10 minutes?) and return `n`. This would be preferable to the job getting stuck on this request and continuing to block nodes.

fkraushofer (commented):
Looks good overall. One point to discuss concerning how to handle user confirmations

src/viperleed/calc/bookkeeper/utils.py
michele-riva: Good, point. I had not considered non-interactive shells. Perhaps we could even detect whether we have the option of asking user input in the first place, and `BookkeeperExitCode.FAIL` if this is not the case? We can then suggest to run `bookkeeper` manually in an interactive shell.

By chance, do you know what does `sys.stdin.isatty()` return in a slurm-handled session?

michele-riva: I have implemented a solution based on `sys.stdin.isatty()` in a41629d3e6ee14e2f2143836586208dcb2ea715a. I'd ask you to still check that `sys.stdin.isatty() == False` in a slurm session, as I have little experience with it. Probably submitting a job that runs a simple script like this one would do the trick:

```python
import sys
print("Interactive?", sys.stdin.isatty())
```

michele-riva: Out of curiosity, it would also be nice to know whether python's `input` function raises in a non-interactive shell. Maybe one could submit another simple script like this one:

```python
"""Test interactive input in a non-interactive shell."""
print(input('This may raise some exception for non-interactive shells'))
```

and have slurm kill it after a short time?

fkraushofer: > Out of curiosity, it would also be nice to know whether python's `input` function raises in a non-interactive shell. Maybe one could submit another simple script like this one:
> 
> ```python
> """Test interactive input in a non-interactive shell."""
> print(input('This may raise some exception for non-interactive shells'))
> ```
> 
> and have slurm kill it after a short time?

When run via slurm, this indeed returns `Interactive? False`

fkraushofer: > Out of curiosity, it would also be nice to know whether python's `input` function raises in a non-interactive shell. Maybe one could submit another simple script like this one:
> 
> ```python
> """Test interactive input in a non-interactive shell."""
> print(input('This may raise some exception for non-interactive shells'))
> ```
> 
> and have slurm kill it after a short time?

Output:
```
This may raise some exception for non-interactive shellsTraceback (most recent call last):
  File "/home/fs70844/surflo/tests/interactive-input/test.py", line 2, in <module>
    print(input('This may raise some exception for non-interactive shells'))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
EOFError: EOF when reading a line
```

fkraushofer: So, the solution should work. I would probably still advocate for some override-mode (yes-to-all) flag that can be passed though, both to bookkeeper and to calc (to pass along to bookkeeper). It's nice to have in general, but also if we ever miss some case where it asks for user confirmation even though it shouldn't, then you would effectively be unable to run bookkeeper in non-interactive shells. Always good to have a fallback in case of such bugs.

src/viperleed/calc/bookkeeper/utils.py
```diff
@@ -15,12 +15,21 @@
...
 
 
-def ask_user_confirmation():
-    """Request user input on an action."""
+def ask_user_confirmation(mode):
+    """Request user input on an action when bookkeeper runs in `mode`."""
+    if not sys.stdin.isatty():
+        LOGGER.error('Bookkeeper needs to ask your confirmation, but '
+                     'this shell is not interactive. Please run again '
+                     f'\'bookkeeper {mode.long_flag}\' in an interactive '
```
fkraushofer, line 30: Nitpick: Should be 'Please run bookkeeper again'

src/viperleed/calc/bookkeeper/utils.py
amimre: I think that is the same issue as with using other parameters in the bookkeeper – PARAMETERS is only read after the first bookkeeper execution.

michele-riva: I can add a `-y` CLI arg to `calc` as well, and pass it on to `bookkeeper` when called by `calc`. While I'm not a huge fan, we can also reuse the `-y` CLI arg of bookkeeper for the other user-confirmation questions. Concerning `HALTING`, as @amimre mentioned it is indeed the same (philosophical) problem of not having bookkeeper deal with any of the contents of the input files. However, we may eventually have to do this anyway to correctly handle #215.

michele-riva: Done in 4c92ff5488d1755b23082e8eaf309ed35543e1ee + 64f44b5cfcdf209eecd9100e61ceded9143de32f

michele-riva:
@fkraushofer: I have implemented things as discussed above. Would you mind taking another look?

src/viperleed/calc/bookkeeper/cli.py
```diff
@@ -110,8 +110,8 @@ def add_parser_arguments(self, parser):
             )
         parser.add_argument(
             '-y',
-            help=('Do not ask for confirmation when running in '
-                  f'{BookkeeperMode.DISCARD_FULL.long_flag} mode.'),
+            help=('In interactive shells, automatically reply "yes" '
+                  'to all requests for user confirmation.'),
```
fkraushofer, line 114: we would want this to also be the case in non-interactive shells, I suppose?

src/viperleed/calc/bookkeeper/bookkeeper.py
```diff
@@ -106,17 +106,27 @@ def tensor_number(self):
...
         ----------
         mode : str or BookkeeperMode
             Which bookkeeper mode to use. See help(BookkeeperMode).
         requires_user_confirmation : bool, optional
-            Whether user confirmation is necessary before proceeding
-            with destructive actions. Only used in DISCARD_FULL mode.
-            Default is True.
+            Whether user confirmation is necessary before proceeding.
+            Only used if the current shell is interactive. If not, the
+            automatic action is the same as if the user replied "no".
```
fkraushofer, line 119: So does this mean there's still no way to override in non-interactive shells? The whole point of the `-y` tag to me was that if I schedule a run in a non-interactive shell, and I already know there will be a request for confirmation (due to a bug, or something weird about my folder structure that I'm aware of but don't want to change), I can preemptively override, instead of having to live with the bookkeeper just failing every time

fkraushofer (commented):
Either the `-y` tag does not do what I wanted, or the documentation is incorrect; I did not dive in to try and figure out which. See below.

michele-riva:
> Either the `-y` tag does not do what I wanted, or the documentation is incorrect; I did not dive in to try and figure out which. See below.

Sorry, I think the problem is indeed with the documentation. I will add some tests to confirm that this is the case.

src/viperleed/calc/bookkeeper/cli.py
michele-riva: Was a documentation problem. Fixed in 02f039153f578474512717e5e842adfccaa8ce97

src/viperleed/calc/bookkeeper/bookkeeper.py
michele-riva: Was a documentation problem. Fixed in 02f039153f578474512717e5e842adfccaa8ce97

michele-riva:
@fkraushofer 

I can confirm that it was a documentation issue. Fixed now.

fkraushofer (approved):

Issue #185, opened 2024-03-06
Accept multi-line SITE_DEF


alexmiame:
It seems like in the latest version, multi-line SITE_DEFs for the same POSCAR Element are broken.

Using e.g.
`SITE_DEF A = surf 5 6, sub_1 3 4`
works, but with
```
SITE_DEF A = surf 5 6
SITE_DEF A = sub_1 3 4
```
 the first site definition is ignored/overwritten, and no `A_surf` sites are assigned.

PR #332, opened 2025-04-17
Fix TypeError, more SITE_DEF tests

amimre:
### Edit @michele-riva:

After discussion, we decided to keep the `SITE_DEF` behavior unchanged. Thus, the description below is outdated. This PR fixes a `TypeError` in `superpos`, and adds some tests for multiple assignment to a `SITE_DEF`.

### Original, from @amimre

Hi all,
I just noticed a small but rather important bug in the parsing of PARAMETERS. Counter to what we intended, giving multiple `SITE_DEF`s on multiple lines for the same POSCAR element would silently overwrite each other. This PR fixes that and adds a regression test.
I would suggest to make sure to include this fix in `v0.13.0`.


amimre:
I'm also grouping in a syntax error I just noticed in superpos.py

michele-riva:
I think we had discussed this already in #185. The expected (and documented) syntax for `SITE_DEF` is

```
SITE_DEF <poscar_element> = <list of definitions>
```

Essentially, repeating `SITE_DEF` with the same "flag" more than once is equivalent to overwriting, as it is for (almost) all the other parameters. I believe this is the better way to deal with this, as having multiple `SITE_DEF`s for the same POSCAR element is likely a user "error". So I'd prefer to leave the current behavior unchanged. I'd like @fkraushofer's opinion on this too, though.

amimre:
Huh, I completely forgot about that!
Seems it's not quite as intuitive (at least for me).

I'm setting up a system with layer-dependent chemistry, for which I was using this syntax:
```
SITE_DEF Me = layer_1 3-5
SITE_DEF Me = layer_2 6-8
SITE_DEF Me = layer_3 9-11
SITE_DEF Me = layer_4 12-14
```
Putting it all in one line is possible, but maybe a bit unwieldy.
If we keep the current behavior, I would at least warn the user about overwriting the previous assignment.

michele-riva:
Concerning warning: Sure we can do that, but then we should do so for all the other `PARAMETERS` that we also silently ignore. I would not treat `SITE_DEF` differently than any others. Doing all the checks right seems like a lot of work, as sometimes it is enough to have a repeated `"PARAMETER = ..."` line to overwrite `PARAMETER`, sometimes repeating `"PARAMETER flag = ..."` causes overwriting, sometimes neither (e.g. `DOMAIN`).

michele-riva:
> Huh, I completely forgot about that! Seems it's not quite as intuitive (at least for me).
> 
> I'm setting up a system with layer-dependent chemistry, for which I was using this syntax:
> 
> ```
> ELEMENT_MIX Me = Rh, Pt
> SITE_DEF Me = layer_1 3-5
> SITE_DEF Me = layer_2 6-8
> SITE_DEF Me = layer_3 9-11
> SITE_DEF Me = layer_4 12-14
> ```
> 
> Putting it all in one line is possible, but maybe a bit unwieldy. If we keep the current behavior, I would at least warn the user about overwriting the previous assignment.

I believe the rationale behind this original choice was to avoid users "forgetting" some `SITE_DEF` lines somewhere else in their PARAMETERS file. Having all definitions on one line (for one POSCAR element) ensures that we correctly interpret the users' will, by forcing them to look at the whole line.

michele-riva:
I'll close this for now, we can open it again after hearing from @fkraushofer if needed.

@amimre: could you please
1. Create a separate branch with the changes in this PR, then revert the commits in `bugfix/calc`? This way we keep the latter "clean" for future PRs.
2. Make a separate PR for the `TypeError` in superpos?

amimre:
I would maybe not fully revert everything and instead only undo the changes to `interpret.py` and add the adapted test case anyhow (i.e. multiple SITE_DEFS, but defined on one line) since we didn't have it before.
How does that sound?

michele-riva:
Definitely, good idea. Always nice to have more tests.

fkraushofer:
Hey guys, sorry for the late response - busy days.

My 2 cents: both allowing and not allowing definitions on multiple lines has pros and cons, as you have already discussed. Allowing multi-line definitions can make things for readable especially for complex systems, but introduces some potential for errors.

I have no strong opinion on either one, but agree it should be consistent for all parameters.

src/viperleed/calc/sections/superpos.py
```diff
@@ -227,8 +227,8 @@ def _best_config_from_sdtl(rp):
     # check for an SD.TL file
     cwd = Path()
     sdtl_files = (
-        f for f in (cwd/"SD.TL", cwd/{DEFAULT_OUT}/"SD.TL") if f.is_file()
-        )
+        f for f in (cwd/"SD.TL", cwd/f"{DEFAULT_OUT}/SD.TL") if f.is_file()
+    )
```
michele-riva, line 231:
```suggestion
        )
```

Indent style: closing parenthesis aligned with previous line.

tests/calc/files/parameters/case_parameters.py
```diff
@@ -62,6 +63,14 @@ def _get_poscar_info(name):
            'BULK_LIKE_BELOW': 0.35, 'T_DEBYE': 420, 'T_EXPERIMENT': 100,
            'SITE_DEF': {'Ir': {'surf': {3, 2}}, 'O': {'ads': {1}}},
            'VIBR_AMP_SCALE': ['*surf 1.3',], 'V0_IMAG': 5.0},
+    'PtRh': {'RUN': [0],
+           'THEO_ENERGIES': [49, 550, 3], 'ELEMENT_MIX': {'Me': ['Rh', 'Pt']},
+           'BULK_LIKE_BELOW': 0.45, 'T_DEBYE': 250, 'T_EXPERIMENT': 300,
+           'SITE_DEF': {
+               'O': {'surf': {1, 2}},
+               'Me': {'layer_1': {3, 4, 5}, 'layer_2': {6, 7, 8},
+                      'layer_3': {9, 10, 11}, 'layer_4': {12, 13, 14}}},
```
michele-riva, line 72:
```suggestion
           'SITE_DEF': {
               'O': {'surf': {1, 2}},
               'Me': {'layer_1': {3, 4, 5}, 'layer_2': {6, 7, 8},
                      'layer_3': {9, 10, 11}, 'layer_4': {12, 13, 14}},
               },
```

Indent style: closing bracket on multi-line is on its own line.

tests/calc/files/parameters/case_parameters.py
```diff
@@ -62,6 +63,14 @@ def _get_poscar_info(name):
...
+           'VIBR_AMP_SCALE': ['*surf 1.3',]},
```
michele-riva, line 73:
```suggestion
           'VIBR_AMP_SCALE': ['*surf 1.3',],
           },
```

Same as above

tests/calc/files/parameters/case_parameters.py
```diff
@@ -62,6 +63,16 @@ def _get_poscar_info(name):
            'BULK_LIKE_BELOW': 0.35, 'T_DEBYE': 420, 'T_EXPERIMENT': 100,
            'SITE_DEF': {'Ir': {'surf': {3, 2}}, 'O': {'ads': {1}}},
            'VIBR_AMP_SCALE': ['*surf 1.3',], 'V0_IMAG': 5.0},
+    'PtRh': {'RUN': [0],
```
michele-riva, line 66:
```suggestion
    'PtRh': {
        'RUN': [0],
```

Sorry, one more I just noticed.

michele-riva (changes requested):
Sorry, one last small style change.

tests/calc/files/parameters/case_parameters.py
```diff
@@ -62,6 +63,17 @@ def _get_poscar_info(name):
            'BULK_LIKE_BELOW': 0.35, 'T_DEBYE': 420, 'T_EXPERIMENT': 100,
            'SITE_DEF': {'Ir': {'surf': {3, 2}}, 'O': {'ads': {1}}},
            'VIBR_AMP_SCALE': ['*surf 1.3',], 'V0_IMAG': 5.0},
+    'PtRh': {
+        'RUN': [0],
+           'THEO_ENERGIES': [49, 550, 3], 'ELEMENT_MIX': {'Me': ['Rh', 'Pt']},
```
michele-riva, line 68: Sorry, again: these other keys should be aligned with the previous one.

michele-riva (changes requested):
I know, I promised... This is the last one. Please have al the keys aligned to the same column (as well as the closing bracket).

amimre:
Yeah, this is why we should really implement auto-formatting...
(*cough* #276 *cough*)

michele-riva (approved):
Good to go!

PR #347, opened 2025-05-15
Fix links to documentation in README

michele-riva:
Fixes some issues with links to the documentation in the README.md file.

@amimre: I somehow can't find any longer the "latest" build on readthedocs, and the current default changed to "stable". I tried to pull it back or renaming "stable", but did not manage. Perhaps we should sit together for a quick look?


amimre (approved):
Looks good, though both `www.viperleed.org` and `viperleed.org` work.
(The former is the canonical address, and the latter forwards to it.)

PR #349, opened 2025-05-15
fix: Reset search_index after all blocks are done

fkraushofer:
This fixes an issue with the search loop logic. Currently, if you set something like `RUN = 1-3 1-3 1`, the first `1-3` will go through all the blocks in `DISPLACEMENTS` as intended. However, when the section loop arrives at the second `1-3`, it will attempt to read a displacement block that's not there any more and terminate with an index error.


michele-riva (approved):
Looks good, but I'd suggest to always merge into the most recent version branch (0.13.1 now) rather than `master`. This way, `master` always contains the most recent released code. I've already changed the target branch. We can release this together with the rest (today, I guess).

Issue #344, opened 2025-05-13
Bookkeeper deletes VIBROCC file from root directory after full-dynamic optimization


fkraushofer:
I'm running v0.13.0, installed via pip.

After a full-dynamic optimization (with otherwise default settings), the root directory contains a `VIBROCC_ori`, but no more `VIBROCC`. Likewise, `OUT` contains a `POSCAR`, but no `VIBROCC`. 

Changing nothing and simply proceeding to the next job (another full-dynamic optimization in my case) caused the following behavior in the next run:
- when the job starts, the root directory is initially unchanged; `VIBROCC_ori` is still present.
- `calc` fails immediately because no `VIBROCC` is present in `work`
- in the bookkeeper run at the end of the job, the `VIBROCC_ori` is now also deleted (!). No `VIBROCC` of any sort is now present in the root directory.
- The `OUT` directory is empty
- There is now no more `POSCAR` in the root directory, only a `POSCAR_ori`. Continuing from here causes the same issues with `POSCAR`, i.e. the run immediately fails and the bookkeeper deletes the `POSCAR_ori` as well.

Obviously, it's possible to restore from `history`, but this is still a breaking bug. @michele-riva could you prioritize this and update us to a v0.13.1 asap?

The original cause here appears to be that the full-dynamic optimization does not create a `VIBROCC` in `OUT`. However, the followup errors reveal a more fundamental issue, which is that the bookkeeper deletes stuff without replacement whenever no `OUT` files are generated for whatever reason (e.g. a run that fails early).

michele-riva:
Hmm I see. I would like to test this with the directory structure produced in the various steps and make sure all work as expected. Could you perhaps attach here the original input files with which you started the first FD optimization?

michele-riva:
Having your `bookkeeper.log` and `calc` logs would also be useful.

fkraushofer:
I uploaded the whole folder including history to our cloud.
Alternatively on VSC4: `/gpfs/data/fs71704/fkraushofer/viperleed_calc/Cu2O-111/1x1/CuD`, you should have access.

michele-riva:
Thanks, I'll look into it.

michele-riva:
[...]

PR #351, opened 2025-05-15
v0.13.2

### LINKED ISSUE #353: No `SUPP/original_inputs` for very early `calc` failure
When `run_calc` exits very early because of:
* A failure to read/interpret `PARAMETERS`
* A failure to read/interpret `POSCAR`
* An early `HALTING` condition

the original input files are not collected into `SUPP/original_inputs`: https://github.com/viperleed/viperleed/blob/b9e4290e6f794f39023e8e91809acff28eb28be9/src/viperleed/calc/run.py#L118-L142

`preserve_original_inputs` is called after early-exit conditions are sorted out. Here the current implementation of `_finalize_on_early_exit`: https://github.com/viperleed/viperleed/blob/b9e4290e6f794f39023e8e91809acff28eb28be9/src/viperleed/calc/run.py#L150-L155

This issue is related to #344 in that the outcome from the point of view of `bookkeeper` is similar: since there are neither `OUT` nor `SUPP/original_inputs` files, the root directory after `bookkeeper` executes only contains `*_ori`-suffixed files. In fact, after suffixing those in root, the unlabeled files are pulled (in order) from either `OUT` or `SUPP/original_inputs`.

I can see two courses of action here:
1. Modify `_finalize_on_early_exit` to also call `prerun_clean` (in case there is any previous `calc` results) and `preserve_original_inputs`. This will
    * store the (likely faulty) `SUPP/original_inputs` files
    * make `bookkeeper` treat this case correctly
    
    However, it would require to adapt `prerun_clean` and `preserve_original_inputs` to handle the case in which no `rpars` exists.
2. Treat this case differently in `bookkeeper`, likely with a dedicated method that should be called in `calc.cli.ViPErLEEDCalcCLI.__call__` in case `run_calc` exits early. The method would simply duplicate all `*_ori`-suffixed files to their unlabeled version, as long as no unlabeled version is present in root. This case would bear the obvious disadvantage that, should anything fail before or during this method, the next invocation of `bookkeeper` (e.g., the `--clear` that happens automatically before `calc`) would not be able to correctly discern this situation. Also, we currently (correctly) exit with a no-failure condition in the `HALTING` case, so discerning these situations would be complicated.

I would probably be more in favor of option 1, and will start working on an implementation. @fkraushofer, @amimre: let me know if you think option 2 is better, and whether you see any problem with option 1.

**Edit**: notice that those listed above are the only situations in which we exit "cleanly" from `calc`, i.e., without raising an exception. There can be other situations that make a call to `viperleed.calc` fail:
* Creation of the work directory fails (typically `OSError`)
* Copying Tensors/Deltas to work fails (except for `FileNotFoundError` in the `all_tensors` case)
* Copying input files to work fails (except for `FileNotFoundError`)

These situations raise exceptions and will thus prevent the `bookkeeper` call after `calc` from happening. However, they should be safe, as there is no `calc` log file yet, so the input directory is not considered to be archived. Thus, the `bookkeeper` call before the next `calc` will not do anything.

michele-riva:
Fixes #353 and the bookkeeper-related issue pointed out by @fkraushofer in https://github.com/viperleed/viperleed/issues/344#issuecomment-2884404676.

I would release this only with this fix considering that several people may be affected.

### DIFF
--- a/src/viperleed/__init__.py
+++ b/src/viperleed/__init__.py
@@ -11,7 +11,7 @@ __authors__ = (
 __copyright__ = 'Copyright (c) 2019-2025 ViPErLEED developers'
 __created__ = '2020-08-03'
 __license__ = 'GPLv3+'
-__version__ = '0.13.1'
+__version__ = '0.13.2'
 
 
 GLOBALS = {
--- a/src/viperleed/calc/bookkeeper/bookkeeper.py
+++ b/src/viperleed/calc/bookkeeper/bookkeeper.py
@@ -18,6 +18,7 @@ from viperleed.calc.bookkeeper import log
 from viperleed.calc.bookkeeper.domain_finder import DomainFinder
 from viperleed.calc.bookkeeper.domain_finder import MainPathNotFoundError
 from viperleed.calc.bookkeeper.errors import _FileNotOlderError
+from viperleed.calc.bookkeeper.errors import BookkeeperUnexpectedError
 from viperleed.calc.bookkeeper.errors import NotAnInteractiveShellError
 from viperleed.calc.bookkeeper.exit_code import BookkeeperExitCode
 from viperleed.calc.bookkeeper.history.constants import HISTORY_INFO_NAME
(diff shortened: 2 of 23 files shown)


michele-riva:
Thanks @fkraushofer for approving #354. This one has only changes from there, plus the version bump. Would you mind reviewing?

fkraushofer (dismissed):

michele-riva:
In the meantime, #356, #357, and #358 are awaiting approval/have been approved already. So I'd have these go to 0.13.2 as well.

michele-riva:
@fkraushofer : this is good to go from my point of view. Would you mind reviewing?

michele-riva:
@fkraushofer @amimre 

Not sure you got a notification already, but 0.13.2 is out now. Docs are also built.

PR #356, opened 2025-05-16
Minor change to disable a nagging info log message

fkraushofer:
I've noticed that we constantly report that Search-progress.csv has been written while the search is running. This is inconsistent with all other file creations, which are debug-messages at most, and not even that for the minor files that get written continuously during the search.


michele-riva (approved):
Looks good to me! Thanks for removing the annoying repeated logs.

PR #359, opened 2025-05-18
Update license information in pyproject.toml

michele-riva:
Adapt license information to the most recent standard (PEP 639, see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license). Fixes `DeprecationWarnings` in `setuptools`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,8 @@ name = 'viperleed'
 description = 'Tools for quantitative low-energy electron diffraction'
 requires-python = '>=3.7'
 dynamic = ['version']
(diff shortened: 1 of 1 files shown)


fkraushofer (approved):

michele-riva:
Turns out I was too zealous in doing this. PEP639 support has been added in setuptools only with version 77.0.3 (see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files), which does not support python 3.7. I will undo the changes while we figure out how to handle this in a backward-compatible manner.

PR #369, opened 2025-05-26
Note on escaping brackets for optional dependencies in zsh

amimre:
Adds a short note on how to escape the brackets used in e.g. `pip install viperleed[GUI]` which are interpreted as a glob pattern in the zsh shell.


michele-riva:
@amimre: Should I re-review?

michele-riva (approved):
Looks good!

Issue #74, opened 2023-02-21
GUI - graphical pattern output


fkraushofer:
It would be very helpful if the GUI could export the currently displayed pattern as a PDF file. This could then be used to overlay on experimental LEED patterns, and/or for publication.

It would also be nice if there was an option to show all spots in the same color when there are multiple domains. Color-coding domains is often helpful, but can make it harder to compare patterns by eye.

PR #371, opened 2025-05-27
Refactor guilib to gui

michele-riva:
- Renames `guilib` to `gui`
- Reworks the `gui` package to conform with the layout of other `viperleed` subpackages, i.e., with `cli.py` and `__main__.py` modules. Also moves authoring and created-date information to module dunders. Adds licence and copyright information on all modules.
- Adds the `--nogui` CLI argument via `argparse`.

TODO: add tests for `cli.py` module.


amimre:
What is the reasoning behind the `--no-gui` option? Do you want to enable GUI only features (LEED pattern simulation?) without GUI?

michele-riva:
> What is the reasoning behind the `--no-gui` option? Do you want to enable GUI only features (LEED pattern simulation?) without GUI?

That was an old suggestion by @fkraushofer (which is still not implemented). Would allow accessing the pattern simulator without a GUI interface. Ideas of what this could be used for:
- exporting a pattern file given an input
- exporting a figure of the simulated pattern (not possible yet even in the GUI, see #74)

I would not worry much about the exact functionality at this stage, nor the CLI arg itself, as it is simply not implemented. Once it is, we can rethink how it could work.

amimre (approved):

PR #373, opened 2025-05-27
Replace quote character in the header of spot pattern files

### LINKED ISSUE #86: Quote character in pattern file header
@schmid-iap pointed out that having double quote marks `"` in the header of the .csv pattern files seems to be problematic when opening with Excel. It seems that they are somewhat treated as special characters and escaped funnily.

Simple solution: use single quotes.

michele-riva:
Fixes #86.


michele-riva:
Since I know that @amimre is busy, and considering the small changes that this one implements, I go ahead and will merge it. This way I can proceed with the rest. @amimre: let me know if you feel any changes are needed.

PR #374, opened 2025-05-27
Update information in about dialog

michele-riva:
See 917e2aaa168ba05d15c83711b74c4af2b9228286.

Issue #318, opened 2025-04-03
Graphics check in posix defunct


FlorianDoerr:
It seems that the check whether a mac has a graphical display (in the guilib `__init__`) falsely seems to return that there is no graphical display. (on @SinTu404's laptop)

michele-riva:
@amimre: could you please check whether you can start the gui on your machine (use branch `measurement-to-separate-thread`)? Or does it simply say "running in command-line mode. Not implemented yet"?

michele-riva:
Update: the check that used to be working fine on my WSL seems to also be not doing the right thing there any longer. It appears that checking whether a `$DISPLAY` environment variable is present is not enough to guarantee that a display actually exists and can be connected to.

We will likely need to implement a (platform-dependent?) check of whether a display is available. An alternative would be to check whether Qt5 provides any platform-agnostic way to detect whether a display is available. In fact, at the point where the abovementioned check is executed, we should already know whether PyQt5 is installed on the system.

michele-riva:
I upped this to `PRIORITY` as we need this for the upcoming ViPErLEED workshop

FlorianDoerr:
My understanding is that with #363 the `ImportError` will be caught and swallowed without further warning in the `__init__` of the camera module. In the future we should report something like "Camera type x could not be imported and will therefore not be detected."

michele-riva:
> My understanding is that with #363 the `ImportError` will be caught and swallowed without further warning in the `__init__` of the camera module. In the future we should report something like "Camera type x could not be imported and will therefore not be detected."

Good idea

michele-riva:
A quick test on Ubuntu and Mac reveals that now this is fixed (at least to the extent that the GUI can be started).

On Ubuntu, however, we had to

```
sudo apt install libxcb-xinerama0
```

following suggestions under https://forum.qt.io/topic/93247/qt-qpa-plugin-could-not-load-the-qt-platform-plugin-xcb-in-even-though-it-was-found/57. Otherwise got the following Qt error message (with corresponding crash of the GUI):
```
Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

[...]

Issue #211, opened 2024-06-27
Pattern simulator is broken


michele-riva:
Currently none of the branches in this repository contains a working version of the pattern simulator. We have discussed with @fkraushofer and @amimre that, for v1.0.0, it is sufficient for me to fish back a working version from the early days, probably the one that was packed to the viperleed-betatest repository.

I'm making this issue just as a reminder for myself.

PR #378, opened 2025-05-29
Reactivate Pattern Simulator GUI

### LINKED ISSUE #9: mplcairo + matplotlib crash GUI with python 3.8
When using matplotlib with the mplcairo backend on python 3.8 x64 the GUI crashes suddenly when loading some of the input files (e.g., Cu(111)-(3xr3)rect.tlm, while Cu(111)-(1x1).tlm loads but the LEED pattern plots weird stuff).

**Tested on:**
1. win7-x64; pyw3.8-x64 --> **BUG IS THERE**
  * Matplotlib: v3.4.1
  * mplcairo: v0.4
2. win7-x64; py3.7-x86  --> **NO BUG**
  * Same modules version as above

It could well be that the problem is with the cairo .dll x64 rather than with the modules themselves, as the crash does not lead to any traceback exception, but simply solidly crashed the interpreter at the OS level.

**To be tested on a different machine**

Right now, I will temporarily solve the issue by picking a different backend when using py3.8. This way we should e able to give a stable release running under py 3.8.

michele-riva:
Fixes #9, #86, #211. #318.

Also collects changes from:
- #371
- #373 
- #374

Notes:
1. No changes should be introduced for what concerns `calc`, at least to the extent that is currently covered in our tests
2. The pattern simulator UI is (essentially) at the same state as it is on the viperleed-betatest page, except that:
    - it starts up by showing the plugin-selection window 
    - the pre-release notice is removed
    - some small bugs were fixed (to the extent that was visible with `pylint --errors-only`)
    - it can run on any platform (on which Qt runs)
3. As a result of point 2, do not expect the GUI to necessarily look nice on all platforms/screens. The basic functionality that should work is (i) loading a file or creating a new input, (ii) showing the LEED pattern, and (iii) exporting a spot-pattern file.


michele-riva:
@amimre @fkraushofer 

I think this is good to go. Unfortunately, with the many filename changes it is not immediate to review from the "Files changed" tab.

What is still missing (next on my list, in #210): `[PatternInfo].tlm` -> `[experiment_symmetry].ini`

michele-riva:
Great, thanks. I agree that we should have a top-level `lib` package at some point. I'd prefer that over having things in `calc.lib`.

fkraushofer:
Please don't wait for my review, won't get to it before some time tomorrow and am a bit tight on time anyway. If you both agree it's good to go, then I'm sure it's fine.

Issue #377, opened 2025-05-28
Camera widget for exposure time raising error


FlorianDoerr:
Setting the exposure time on the camera can potentially break the camera viewer if all remaining digits in the QLineEdit are zeros. I will check if setting a minimum time (e.g.: of 1 ms) is enough to fix the issue.

michele-riva:
We should use your soft range line edit. The min and max are set according to what rhe camera can do.

michele-riva:
I've created #390 to fix this and improve the user experience.

michele-riva:
Fixed in #390. Closing now as we will not merge it in master yet, but whenever we release the measure module.

PR #387, opened 2025-05-31
Port src layout to settings dialog

michele-riva:
Fixes all current merge conflicts in #263 with the measurement-to-separate-thread base branch that resulted from moving to a `src` layout.

@FlorianDoerr: we should first check and merge #386, as this one contains changes from there too. Then, the number of changes will be limited.

Please still hold on to your uncommited changes till this is merged.


FlorianDoerr:
I would assume that a minimum exposure time of 0 ms may lead to some strange behaviour down the line when we have other cameras implemented, but I think managing this kind of issue is going to be a matter of subclasses. I see no further issue with this right now.

FlorianDoerr (approved):

FlorianDoerr:
> I would assume that a minimum exposure time of 0 ms may lead to some strange behaviour down the line when we have other cameras implemented, but I think managing this kind of issue is going to be a matter of subclasses. I see no further issue with this right now.

On second thought, since setting an exposure time of 0 ms will in fact break the imagingsource camera, we should not allow this.

michele-riva:
> > I would assume that a minimum exposure time of 0 ms may lead to some strange behaviour down the line when we have other cameras implemented, but I think managing this kind of issue is going to be a matter of subclasses. I see no further issue with this right now.
> 
> On second thought, since setting an exposure time of 0 ms will in fact break the imagingsource camera, we should not allow this.

I think this is not a problem of having zero here. The problem is that we try to set an exposure that is smaller than the minimum. If we were to have another camera whose minimum is 2ms, it's clear (and correct) that trying to set a smaller value (e.g. 1ms) leads to problems.

michele-riva:
Update: As I mentioned in #377, I cannot reproduce the problem in this branch, so I think using a `CoercingDoubleSpinBox` as we do now should not give any problem with `CameraABC.get_exposure_limits()` returning `(0, np.inf)`.

Issue #395, opened 2025-06-03
Single-threaded search fails


fkraushofer:
I just tried running a search with `N_CORES = 1`, which should start the single-threaded search. This crashes with a banal-seeming python error. Should be an easy fix:

```
# WARNING: The N_CORES parameter is set to 1. The search will be run without parallelization. This will be much slower!
Compiling fortran input files...
Search log will be written to file search-250603-151740.log.
Starting search. See files Search-progress.pdf and SD.TL for progress information.
### ERROR ### in run_sections:run_section:226
# Error in section SEARCH
#############
### ERROR ### in run_sections:section_loop:381
# Exception during viperleed.calc execution: 
#############
Traceback (most recent call last):
  File "/home/fs70824/trainee124/source/viperleed/src/viperleed/calc/sections/run_sections.py", line 278, in section_loop
    run_section(sec, sl, rp)
  File "/home/fs70824/trainee124/source/viperleed/src/viperleed/calc/sections/run_sections.py", line 218, in run_section
    search.search(sl, rp)
  File "/home/fs70824/trainee124/source/viperleed/src/viperleed/calc/sections/search.py", line 964, in search
    command.append(executable)
    ^^^^^^^
UnboundLocalError: cannot access local variable 'command' where it is not associated with a value
```

PR #399, opened 2025-06-05
Fix: invalid bulk3dSym when saving experiment-symmetry file

michele-riva:
Also
- updates menus/toolbar/tooltips to better reflect the new file-naming convention (and not only extension)
- refactors some Qt widget-related modules into `gui/widgets`


michele-riva:
Not sure if we can make this happen for the workshop. If so, we could also do #395.

PR #422, opened 2025-06-09
 Document how to handle failure to install GUI on missing cairo dependencies

### LINKED ISSUE #415: Install GUI fails on Mac & py3.13
Today at the LEED-I(V) workshop an installation problem came up with @MaxBuchta.

Steps to reproduce:
pip install "viperleed[gui]"

Fails at installing ~~quicktions~~ mplcairo (or, better, it's pycairo dependency).

Details on the configuration:

- MacOS 15.5
- python 13.3.4
- pip 25.1.1

michele-riva:
Fixes #415

### DIFF
--- a/doc/content/installation.rst
+++ b/doc/content/installation.rst
@@ -80,6 +80,31 @@ in your terminal.
         # or
         python -m pip install viperleed\[GUI\]
 
+.. note::
+
+    On Unix and macOS, installation of the ``mplcairo`` package and its
+    ``pycairo`` dependency may fail if either ``pkg-config`` or ``cairo``
+    are not available on the system. If this is the case, you should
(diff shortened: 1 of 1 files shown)


michele-riva:
An alternative to only documenting this would be to add `mplcairo` to an additional `cairo` optional-dependency group. This is because `mplcairo` is only used if present [see https://github.com/viperleed/viperleed/issues/415#issuecomment-2955959162 for details].

Then, people would normally install with
```
pip install viperleed[gui,cairo]
```

but can fall back to
```
pip install viperleed[gui]
```

if the `mplcairo`/`pycairo` installation fails.

amimre (approved):

PR #423, opened 2025-06-10
Do not spam empty folder with empty calc/bookkeeper results

### LINKED ISSUE #408: Skip cleanup/bookkeeper if there are no input files
If `viperleed calc` is called in a completely empty directory, it still creates a bunch of files and directories:
`history  history.info  OUT  SUPP  viperleed-calc-250607-090627.log  work`

This is not really useful, and the mess is a bit annoying when it happens by mistake (happened a few times with the pre-testers in preparation for the workshop).

I would suggest that if not a single `viperleed`-type input file is present at startup, we should produce only the log and nothing else.

michele-riva:
Fixes #408.

Here the necessary steps:
- [x] Do not copy back from work empty `SUPP` or `OUT`
- [ ] Do not execute the second bookkeeper call if no inputs were available
- [ ] Remove `history` and `history.info` if no inputs were available
- [ ] Remove `work` if no inputs were available

### DIFF
--- a/src/viperleed/calc/classes/rparams/rparams.py
+++ b/src/viperleed/calc/classes/rparams/rparams.py
@@ -198,7 +198,7 @@ class Rparams:
         self.halt = 0
         self.systemName = ''
         self.timestamp = ''
-        self.manifest = ManifestFile(DEFAULT_SUPP, DEFAULT_OUT)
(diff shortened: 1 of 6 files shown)


fkraushofer (dismissed):
Looks good. I assume you tested a few actual runs in a "real" application?

PR #333, opened 2025-04-17
Support ViPErLEED JAX plugin

alexmiame:
This PR collects all changes that are required for compatibility with the plugin. To make it more manageable to review, some parts that can stand on their own are split off into separate PRs. These are dependencies that need to be addressed first:
- [ ] #428  (`DISPLACEMENTS` parser)
- [x] #450 (search job class and more robust spawn vs. fork handling)
- [x] #276 (linter; *optional*)
- [ ] #500 (R-factor implementation)

PR #428, opened 2025-06-10
New DISPLACEMENTS parser

alexmiame:
This PR moves the new DISPLACEMENTS file parser from the viperleed_jax repo to the main repository, since it makes more sense to have it here.
It does not (for now) use the new parser to actually read the DISPLACEMENTS file yet.


michele-riva:
Sure, I will. Probably tonight or tomorrow morning.

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,504 @@
...
+    targets_and_direction : str
+        The string to be separated.
+
+    Returns
+    -------
+    tuple
+        A tuple containing the targets string and the direction string. Either
+        string may be empty if the corresponding part was not found.
+    """
+    match = _DIR_AT_END.search(targets_and_direction)
```
michele-riva, line 490: Probably want to ensure there is only one of those.

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+    OccDeltaLine,
+    OffsetsLine,
+    SearchHeaderLine,
+    SectionHeaderLine,
+    VibDeltaLine,
+)
+
+SEARCH_HEADER_PATTERN = re.compile(r'^=+\s+(?i:search)\s+(.*)$')
+SECTION_HEADER_PATTERN = re.compile(
+    r'^=+\s*(OFFSETS|GEO_DELTA|VIB_DELTA|OCC_DELTA|CONSTRAIN)$'
```
michele-riva, line 30: I believe the current syntax requires a single `=`. Of course we can make it more lenient.

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+    GeoDeltaLine,
+    LoopMarkerLine,
+    OccDeltaLine,
+    OffsetsLine,
+    SearchHeaderLine,
+    SectionHeaderLine,
+    VibDeltaLine,
+)
+
+SEARCH_HEADER_PATTERN = re.compile(r'^=+\s+(?i:search)\s+(.*)$')
```
michele-riva, line 28: I think the current syntax requires `==`. We can make it more lenient if there is no risk of confusing this with the sections.

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+        self.current_section = None
+
+    def __next__(self):
+        """Read the next line from the file."""
+        for line in self._file_obj:
+            self._current_line += 1
+            line_content = self._read_one_line(line)
+            if line_content is not None:
+                return line_content
+        raise StopIteration
```
michele-riva, line 72: Do we really need this to override the base-class implementation? We could just raise `ShouldSkipLineError` in `_read_on_line` and let the base class do its job.

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+            self.current_section = None
+            return LoopMarkerLine(LoopMarker.LOOP_START)
+        if LOOP_END_PATTERN.match(stripped_line):
+            self.current_section = None
+            return LoopMarkerLine(LoopMarker.LOOP_END)
+
+        # check for search header
+        search_header_match = re.match(SEARCH_HEADER_PATTERN, line)
+        if search_header_match:
+            label = search_header_match.group(1)
```
michele-riva, line 95: Perhaps give the group a name. It's easier to read.

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+    Note that this class only performs the reading and parsing of the file, but
+    no interpretation. It does not perform any cross-checking and validation of
+    the user input vs. the structure.
+    """
+
+    def __init__(self, filename, noisy=True):
+        """Initialize instance."""
+        super().__init__(filename, noisy)
+
+        self.current_section = None
```
michele-riva, line 63: I would make this private not to risk accessing it an editing from outside.

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+        if section is DisplacementFileSections.OFFSETS:
+            return OffsetsLine(line)
+        if section is DisplacementFileSections.GEO_DELTA:
+            return GeoDeltaLine(line)
+        if section is DisplacementFileSections.VIB_DELTA:
+            return VibDeltaLine(line)
+        if section is DisplacementFileSections.OCC_DELTA:
+            return OccDeltaLine(line)
+        if section is DisplacementFileSections.CONSTRAIN:
+            return ConstraintLine(line)
```
michele-riva, line 129: Perhaps more extensible with a `section_to_line_parser` dict? Or, better, set the Enum values directly to the line parsers rather than auto-assigning them. Then it would just be

```python
if not section:
    raise ...
parser = section.value
return parser(line)
```

src/viperleed/calc/files/new_displacements/reader.py
```diff
@@ -0,0 +1,144 @@
...
+        if search_header_match:
+            label = search_header_match.group(1)
+            self.current_section = None
+            return SearchHeaderLine(label)
+
+        # check for section headers
+        section_header_match = re.match(SECTION_HEADER_PATTERN, line)
+        if section_header_match:
+            section = section_header_match.group(1)
+            self.current_section = section
```
michele-riva, line 103: Would perhaps be good to store directly the DisplacementsFileSection here instead of the string.

src/viperleed/calc/files/new_displacements/tl_backend.py
```diff
@@ -0,0 +1,60 @@
+"""Module tl_backend of viperleed.files.displacements."""
```
michele-riva, line 1: (Part of) this module can probably be moved higher up the calc tree so we will be ready to support more backends, not only search-related. (See my suggestion of having a more generic `BACKEND` parameter in #333.)

src/viperleed/calc/files/new_displacements/tl_backend.py
```diff
@@ -0,0 +1,60 @@
...
+
+__authors__ = ("Alexander M. Imre (@amimre)",)
+__copyright__ = "Copyright (c) 2019-2025 ViPErLEED developers"
+__created__ = "2024-10-15"
+__license__ = "GPLv3+"
+
+class TensorLEEDBackend:
```
michele-riva, line 8: Probably an ABC? Would make it more extensible.

src/viperleed/calc/files/new_displacements/tl_backend.py
```diff
@@ -0,0 +1,60 @@
...
+    """Base class for the tensor LEED backends.
+
+    These classes have information on the capabilities of a given backend, e.g.
+    if it can handle a certain type of search block.
+    """
+
+    def __init__(self, name, handle_search_block_func):
```
michele-riva, line 15: If an ABC, then you can make `replace_search_block` an `@abstractmethod` (can also be a classmethod so we can call it without even making an instance) and avoid passing a callable at creation. The `name` can be made into an abstract property.

src/viperleed/calc/files/new_displacements/tl_backend.py
```diff
@@ -0,0 +1,60 @@
...
+    """Handle the search block with the TensorLEED backend."""
+    # TODO: xy shorthands exits!
+    # For TensErLEED backend, this duplicates a search block xy -> xy[1 0] & xy[0 1]
+    raise NotImplementedError
+
+
+def viplerleed_jax_search_block_handler_func(offsets_block, search_block):
+    """Handle the search block with the VIPERLEED backend."""
+    # no special handling needed, ViPErLEED jax can handle all search blocks
+    return
```
michele-riva, line 51: It should probably return a 1-tuple of the `search_block`?

src/viperleed/calc/files/new_displacements/tl_backend.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self, name, handle_search_block_func):
+        self.name = name
+        self._handle_search_block_func = handle_search_block_func
+
+    def replace_search_block(self, offsets_block, search_block):
+        """Check if the backend can handle the given search block.
+
+        Parameters
+        ----------
+        offsets_block (OffsetsBlock): The offsets block.
```
michele-riva, line 24: I'm not sure I understand the need for the `offsets_block` here.

michele-riva (changes requested):
Looks very very good @amimre. A huge improvement on the current code. I have a few suggestions, and can then have a closer look at the code.

As you will see, one main logical bit seems missing: nested loops.

Of course, the other thing that would need to be implemented is the actual use of the `DisplacementsFile` object with a slab.

src/viperleed/calc/files/new_displacements/tokens/direction.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class DirectionToken(DisplacementsFileToken):
+    """Class to parse direction specifiers in 3D space.
+
+    DirectionToken specifiers are used to declare the allowed directions of
+    movements or constraints for geometric displacements. They can specify one,
+    two, or three dimensions of movements. The direction specifier can be given
+    as one of:
+    - 'xyz'
```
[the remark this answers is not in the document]
amimre: As I mentioned, I don't have the capacity to deal with the parsing of features not needed for the jax version atm, since there's plenty of other things that need doing before.
I don't intend for the new parser to replace the old one immediately anyhow.

src/viperleed/calc/files/new_displacements/tokens/range.py
```diff
@@ -0,0 +1,101 @@
...
+        if self.has_step:
+            return (f'RangeToken(start={self.start}, stop={self.stop}, '
+                    f'step={self.step})')
+        return f'RangeToken(start={self.start}, stop={self.stop})'
+
+def _check_step(step):
+    """Check if the step is positive."""
+    if step is not None and step <= 0:
+        msg = f'Step must be positive: "{step}"'
+        raise RangeTokenParserError(msg)
```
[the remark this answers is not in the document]
amimre: We do?
Maybe we shouldn't since that seems somewhat confusing...

src/viperleed/calc/files/new_displacements/file.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def __init__(self):
+        self.lines = []
+
+    def add_line(self, line):
+        """Add a line to the corresponding section."""
+        self.lines.append(line)
+
+
+class DisplacementsFile:
```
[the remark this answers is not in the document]
amimre: Do we currently properly support nested loops in the DISPLACEMENTS file?
And more importantly, do we want to continue doing so?

src/viperleed/calc/classes/perturbation_type.py
```diff
@@ -0,0 +1,37 @@
...
+
+__authors__ = ("Alexander M. Imre (@amimre)",)
+__copyright__ = "Copyright (c) 2019-2025 ViPErLEED developers"
+__created__ = "2025-05-13"
+__license__ = "GPLv3+"
+
+from enum import Enum
+
+
+class PerturbationTypeError(ValueError):
```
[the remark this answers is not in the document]
michele-riva: Sounds good.

src/viperleed/calc/files/new_displacements/file.py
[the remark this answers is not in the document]
michele-riva: I think we do support them, yes. But we'd need to add some tests to confirm that we indeed do. I think it is worth supporting them, yes. That's how we originally envisioned it. Perhaps @fkraushofer can comment further on this.

src/viperleed/calc/files/new_displacements/tokens/type.py
```diff
@@ -0,0 +1,39 @@
+"""Module for the <type> token in the DISPLACEMENTS file."""
```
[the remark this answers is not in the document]
michele-riva: Sounds good

src/viperleed/calc/files/new_displacements/tokens/base.py
```diff
@@ -0,0 +1,23 @@
+"""Module base of viperleed.files.displacements.tokens."""
+
+__authors__ = ("Alexander M. Imre (@amimre)",)
+__copyright__ = "Copyright (c) 2019-2025 ViPErLEED developers"
+__created__ = "2025-05-12"
+__license__ = "GPLv3+"
+
+from abc import ABC, abstractmethod
+
+class TokenParserError(ValueError):
```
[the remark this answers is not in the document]
michele-riva: Then perhaps we can make it be an `Exception` subclass rather than `ValueError`. We don't want to risk catching it in `except ValueError` blocks.

src/viperleed/calc/files/new_displacements/tokens/direction.py
```diff
@@ -0,0 +1,152 @@
...
+        return self.dof == other.dof and np.allclose(
+            self.vectors, other.vectors
+        )
+
+    def __repr__(self):
+        """Return a string representation of the DirectionToken object."""
+        return f'DirectionToken(vectors={self.vectors}, dof={self.dof})'
+
+
+def _check_unsupported_directions(direction_str):
```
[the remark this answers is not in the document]
michele-riva: Sure. I'll un-resolve this so we keep this in mind.

src/viperleed/calc/files/new_displacements/file.py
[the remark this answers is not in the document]
fkraushofer: Yes, we do - and I think it's worth keeping for the TensErLEED backend. One usecase is an inner loop doing inplane and an outer loop doing z+vib optimization. This can make sense to ensure the inplane is actually converged, and because z+vib tends to be significantly more expensive (because it's a 2D-ish search).

With the new backend, I see no great need for loops in general - but as long as we have them, I think it's not particularly hard to support nesting as well.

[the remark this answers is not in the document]
fkraushofer: Concerning implementation though, I think you may be overthinking it a bit. Correctly treating loops (including nested) only requires a list of the displacement blocks in order, and a list of indices (start, end) for each loop. This is how it's currently implemented. I can give you a quick rundown of how that works in person if you like, since I just recently edited it too.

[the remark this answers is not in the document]
michele-riva: > With the new backend, I see no great need for loops in general - but as long as we have them, I think it's not particularly hard to support nesting as well.

Agreed. We would then just need to decide whether the new backend will "unroll" all loops to construct a larger parameter space, or if we should still keep.

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,504 @@
...
+        """Return the string representation of the line."""
+        return f'{self.targets} = {self.range}'
+
+
+
+class OccDeltaLine(ParsedLine):
+    """Class to parse lines in the OCC_DELTA block of DISPLACEMENTS.
+
+    Lines in the OCC_DELTA block are of the form:
+        <target> [, <target>] = <element> <range> [, <element> <range> ...]
```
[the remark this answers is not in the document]
fkraushofer: I'm happy to deprecate this one. Was never a big fan, and I can't imagine anyone is using it since it's still very complicated to understand the syntax, and as you say, the same can be accomplished by constraints.

src/viperleed/calc/files/new_displacements/file.py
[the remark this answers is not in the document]
amimre: Well, you can't unroll a loop with an undetermined number of iterations, so you need to keep the look information in the object anyhow.
Either way, the parser & reader won't deal with the indices per se.

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,504 @@
...
+    )
+
+    def __init__(self, line: str):
+        super().__init__(line)
+
+        # Left hand side
+        # check if the last part is a direction
+        targets_str, dir_str = separate_direction_from_targets(self._lhs)
+        if dir_str:
+            raise InvalidDisplacementsSyntaxError(self.invalid_format_msg)
```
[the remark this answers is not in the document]
michele-riva: Sorry, I screwed this up. I did not see that the classes had an `.invalid_format_msg` attribute, so I thought that this line would raise an `AttributeError` when executed. I think `DisplacementsSyntaxError` here is fine.

amimre:
Alright, I worked through the issues raised so far.
I really don't have the capacity to work on this any more for the time being, except as far as it is required for the viperleed-jax plugin and the parameter tree, as I really need to work on finishing those for the time being.

michele-riva:
I can have a look at it

tests/calc/files/displacements/test_lines.py
```diff
@@ -0,0 +1,454 @@
...
+                TargetToken('Fe 2'),
+                LinearOperationToken.from_array(np.eye(1)),
+                id='geo-single-target-link',
+            ),
+            pytest.param(
+                'geo A = [1 0 0] B',
+                TypeToken('geo'),
+                [TargetToken('A')],
+                TargetToken('B'),
+                LinearOperationToken('[1 0 0]'),
```
michele-riva, line 364: `LinearOperationTokenParserError: Invalid linear operation format: "[1 0 0]". Expected a 1x1, 2x2, or 3x3 matrix.`

tests/calc/files/displacements/test_lines.py
```diff
@@ -0,0 +1,454 @@
...
+        con = ConstraintLine(line)
+        # Type
+        assert con.type == exp_type
+        # Targets
+        if 'linked' in line:
+            for target, exp_target in zip(con.targets[:-1], exp_targets):
+                assert target == exp_target
+        else:
+            for target, exp_target in zip(con.targets, exp_targets):
+                assert target == exp_target
```
michele-riva, line 405:
```suggestion
        act_targets = con.targets
        if 'linked' in line:
            act_targets = act_targets[:-1]
        assert len(act_targets) == len(exp_targets)
        for target, exp_target in zip(act_targets, exp_targets):
            assert target == exp_target
```

Also, add a comment as to why `[:-1]`.

tests/calc/files/displacements/test_reader.py
```diff
@@ -0,0 +1,33 @@
+"""Tests for files/files/displacements/reader.py."""
+
+__authors__ = ('Alexander M. Imre (@amimre)',)
```
michele-riva, line 3: Some missing dunders. Check also all other test files. Some are missing all of them.

tests/calc/files/displacements/test_reader.py
```diff
@@ -0,0 +1,33 @@
...
+from pytest_cases import fixture
+
+from viperleed.calc.files.new_displacements.reader import DisplacementsReader
+
+
+def _compare_lines(path, expected_lines):
+    with DisplacementsReader(path) as reader:
+        parsed_lines = list(reader)
+    assert len(parsed_lines) == len(expected_lines)
+    for parsed, expected in zip(parsed_lines, expected_lines):
```
michele-riva, line 16: Maybe you want to unpack this `expected` already when iterating into an `(exp_cls, exp_attrs)` tuple?

tests/calc/files/displacements/test_segments.py
```diff
@@ -0,0 +1,59 @@
...
+    ConstraintLine,
+    GeoDeltaLine,
+    LoopMarkerLine,
+    OccDeltaLine,
+    OffsetsLine,
+    SearchHeaderLine,
+    SectionHeaderLine,
+    VibDeltaLine,
+)
+from viperleed.calc.files.new_displacements.segments import SearchBlock, LoopBlock
```
michele-riva, line 19:
```suggestion
from viperleed.calc.files.new_displacements.segments import (
    LoopBlock,
    SearchBlock,
    )
```

tests/calc/files/displacements/test_segments.py
```diff
@@ -0,0 +1,59 @@
...
+        [
+            LoopMarkerLine("<loop>"),
+            SearchHeaderLine("== SEARCH test"),
+            SectionHeaderLine("= GEO_DELTA"),
+            GeoDeltaLine("Fe 1 z = 0.0 0.1"),
+            LoopMarkerLine("</loop>"),
+            LoopMarkerLine("</loop>"),
+        ]
+    )
+    loop_block = LoopBlock(header)
```
michele-riva, line 54: Shouldn't this complain that its only child is another `LoopBlock`?

tests/calc/files/displacements/tokens/test_element.py
```diff
@@ -0,0 +1,41 @@
+"""Tests for <element> token."""
+
+__authors__ = ('Alexander M. Imre (@amimre)',)
+__created__ = '2025-05-13'
+
+# test_element_token.py
```
michele-riva, line 6:
```suggestion
```

tests/calc/files/displacements/tokens/test_offset.py
```diff
@@ -0,0 +1,54 @@
+import pytest
```
michele-riva, line 1: Missing module docstring. Check also other tests, and conform to our standard (see any of the calc tests modules)

tests/calc/files/displacements/tokens/test_offset.py
```diff
@@ -0,0 +1,54 @@
...
+    assert isinstance(token, OffsetToken)
+    assert token.offset == pytest.approx(exp_offset)
+    # repr
+    assert str(token) == f'OffsetToken(offset={token.offset})'
+
+
+def test_init_invalid_count():
+    for bad in ['', '1 2 3 4', 'a b c']:
+        with pytest.raises(OffsetTokenParserError) as exc:
+            OffsetToken(bad)
```
michele-riva, line 34: `@parametrize`; `exc` unused.

tests/calc/files/displacements/tokens/test_offset.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_init_non_numeric():
+    for bad in ['foo', '1.0x', '--3']:
+        with pytest.raises(OffsetTokenParserError) as exc:
+            OffsetToken(bad)
+        assert 'Non-numeric value' in str(exc.value)
```
michele-riva, line 41: `@parametrize`

tests/calc/files/displacements/tokens/test_range.py
```diff
@@ -0,0 +1,112 @@
+"""Tests for module files/displacements/range."""
+
+__authors__ = ('Alexander M. Imre (@amimre)',)
+__created__ = '2025-04-11'
+
+
+import pytest
+
+from viperleed.calc.files.new_displacements.tokens.range import (
+    RangeToken, RangeTokenParserError
```
michele-riva, line 10:
```suggestion
    RangeToken,
    RangeTokenParserError,
```

tests/calc/files/displacements/tokens/test_range.py
```diff
@@ -0,0 +1,112 @@
...
+    # Comparing with non-instance
+    assert dr1 != (0, 1, 0.1)
+
+
+def test_str():
+    dr = RangeToken('1 2')
+    assert str(dr) == 'RangeToken(start=1.0, stop=2.0)'
+
+    drs = RangeToken('1 2 0.25')
+    assert str(drs) == 'RangeToken(start=1.0, stop=2.0, step=0.25)'
```
michele-riva, line 112: Missing newline at end

tests/calc/files/displacements/tokens/test_target.py
```diff
@@ -0,0 +1,126 @@
...
+    # nums and layers
+    assert tok.nums == expected_nums
+    assert tok.layers == expected_layers
+
+
+@pytest.mark.parametrize(
+    'raw',
+    [
+        # combined layer and number
+        'A L(1-2) 3',
```
michele-riva, line 52: Is there a reason why we shouldn't support this?

tests/calc/files/displacements/tokens/test_target.py
```diff
@@ -0,0 +1,126 @@
...
+        # numberic without label
+        '1 2 3',
+        '2',
+        '[1 0 0'
+        # layer without label
+        'L(1)',
+        # wrongly formatted layer
+    ])
+def test_invalid_target_raises(raw):
+    with pytest.raises(TargetingError) as exc:
```
michele-riva, line 62: exc unused

tests/calc/files/displacements/tokens/test_target.py
```diff
@@ -0,0 +1,126 @@
...
+        TargetToken(raw)
+
+
+def test_empty_string_raises():
+    with pytest.raises(TargetingError) as exc:
```
michele-riva, line 67: You can use the `match` keyword of `pytest.raises` rather than checking manually

tests/calc/files/displacements/tokens/test_target.py
```diff
@@ -0,0 +1,126 @@
...
+    a = TargetToken('A 1-2')
+    b = TargetToken('A 1 2')
+    c = TargetToken('A 1-2')
+    # a and c should be equal (same raw string)
+    assert a == c
+    # a and b: same nums but b parsed as nums=[1,2] as well => they compare regex equal & nums equal
+    assert a == b
+    # different label
+    d = TargetToken('B 1-2')
+    assert not (a == d)
```
michele-riva, line 88:
```suggestion
    assert a != d
```

michele-riva: same below

tests/calc/files/displacements/tokens/test_target.py
```diff
@@ -0,0 +1,126 @@
...
+        TargetToken('A L()')
+
+
+def test_generate_label_match_regex_wildcards():
+    # Directly test helper
+    r = _generate_label_match_regex('X*Y')
+    # Should match XANY, XfooYbar, etc.
+    assert r.match('XANY')
+    assert r.match('XfooYbar')
+    assert not r.match('XYX') == None  # ensure it matches "XYX"
```
michele-riva, line 125:
```suggestion
    assert r.match('XYX')
```

michele-riva (changes requested):
Good job @amimre. Mostly small stuff to change.

tests/calc/files/displacements/tokens/test_target.py
amimre: What would be the point?
If I say layers x-y, and atom z it's unclear what the user means.
If atom z is in these layers, does it mean only the atom or everyting? What if the atom is not in those layers?

michele-riva (changes requested):
Sorry @amimre, some more suggestions, but I think we're converging.

michele-riva (changes requested):
Sorry @amimre, some more suggestions, but I think we're converging.

src/viperleed/calc/files/new_displacements/tokens/element.py
```diff
@@ -0,0 +1,54 @@
...
+class ElementToken(DisplacementsFileToken):
+    """Class to parse and represent the <element> token.
+
+    The <element> token in the DISPLACEMENTS file is used to specify chemical
+    (occupational) perturbations and ranges (not to be confused with site
+    labels such as `Fe_surf` used in <target> tokens which may also contain
+    chemical elements). It is used in the OCC_DELTA block and in lines relating
+    to chemical parameters in the OFFSETS and CONSTRAIN blocks.
+    The <element> token is specified by the user as a simple (case-insensitive)
+    string corresponding to the Symbol of a element from the periodic table.
```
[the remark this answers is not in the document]
fkraushofer: Hmm, I think both functionalities are pretty much unused at the moment, and probably noone would miss them if they were gone. 

I can see the use in explicitly defining vacancies, especially if you wanted to limit them to some range - say, you mix 2 elements in arbitrary amounts (which I guess would by default also allow 100% vacancies?), and you want to also set the restriction that at most 10% are vacancies. Sure, you could do that via constraints, but actually, doing it here by saying 0-10% vacancies would be quite intuitive. But also, depending on how it's implemented, I can see that this might be a substantial amount of work... In that case, I'd be fine with making a nice-to-have issue to come back to this later (maybe), and leave it for now.

Defining some element as constant as in the Ti example I find actively confusing and inconsistent with the rest of DISPLACEMENTS. Strongly in favour of deprecating.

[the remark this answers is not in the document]
michele-riva: Fine with me to deprecate the no-stop case of Ti.

@amimre: concerning bounds for vacancies, I believe it would be rather straightforward to implement in JAX: It would probably be just a matter of using the user-defined bounds rather than calculating them out of those of the other elements.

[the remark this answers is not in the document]
amimre: Well, I'd say semi-straightforward – but still work to implement bounds, parsing, tests etc.. I'll table it for now.

src/viperleed/calc/files/new_displacements/tokens/target.py
```diff
@@ -0,0 +1,108 @@
...
+
+class TargetingError(TokenParserError):
+    """Base class for errors in the targeting module."""
+
+
+class TargetToken(DisplacementsFileToken):
+    """Class to handle the <target> token in the DISPLACEMENTS file."""
+
+    def __init__(self, target_str):
+        self.target_str = target_str.strip()
```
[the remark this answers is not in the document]
alexmiame: Strings are parsed at init for other tokens as well.

src/viperleed/calc/classes/perturbation_mode.py
```diff
@@ -0,0 +1,34 @@
...
+    OCC = 'occ'
+    DOM = 'dom'
+
+    def __str__(self):
+        """Return the string representation of the enum."""
+        return self.value
+
+    @classmethod
+    def from_string(cls, mode_string: str):
+        """Convert a string to a PerturbationType enum."""
```
michele-riva, line 29:
```suggestion
    def from_string(cls, mode_string):
        """Convert a string to a PerturbationMode enum."""
```

src/viperleed/calc/files/new_displacements/tokens/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from .base import TokenParserError
+from .direction import CartesianDirectionToken
+from .element import ElementToken
+from .linear_operation import LinearOperationToken
+from .mode import ModeToken
+from .offset import OffsetToken
+from .range_ import RangeToken
+from .target import TargetToken
+from .total_occupation import TotalOccupationToken
```
michele-riva, line 16: Do we need all these init imports? May lead to circular imports later on

src/viperleed/calc/files/new_displacements/tokens/direction.py
```diff
@@ -0,0 +1,205 @@
...
+    ----------
+    direction_str : str
+        The direction string to parse.
+
+    Attributes
+    ----------
+    dof : int
+        The number of degrees of freedom (DOF), either 1, 2, or 3.
+    vectors_xyz : ndarray, shape (dof, 3)
+        A tuple of one, two or three orthonormal vectors spanning the
```
michele-riva, line 80: perhaps here "tuple" is not the correct word, as it is an ndarray

src/viperleed/calc/files/new_displacements/tokens/direction.py
```diff
@@ -0,0 +1,205 @@
...
+        if len(set(direction_str)) != len(direction_str):
+            msg = f'Duplicate directions in direction string: {direction_str}'
+            raise ValueError(msg)
+        if len(direction_str) > 3:
+            msg = f'Direction string too long: {direction_str} (max 3 allowed)'
+            raise ValueError(msg)
+        vecs = [self._get_basis_vector(c) for c in direction_str]
+        return np.array(vecs), len(vecs)
+
+    def _get_basis_vector(self, label):
```
michele-riva, line 145: the three private methods here don't use self and could be `@staticmethod`s

src/viperleed/calc/files/new_displacements/tokens/direction.py
```diff
@@ -0,0 +1,205 @@
...
+        return vec
+
+    def _normalize(self, vec):
+        norm = np.linalg.norm(vec)
+        if norm < DISPLACEMENTS_FILE_EPS:
+            msg = f'Zero-length vector: {vec}'
+            raise ValueError(msg)
+        return vec / norm
+
+    def __eq__(self, other):
```
michele-riva, line 166: These dunders are public, and should appear before private methods

src/viperleed/calc/files/new_displacements/tokens/linear_operation.py
```diff
@@ -0,0 +1,96 @@
...
+        cleaned = op_str.strip()
+
+        # space separated syntax
+        if ',' not in cleaned:
+            # add commas between numbers
+            cleaned = re.sub(r'(?<=\d)\s+(?=\d)', ', ', cleaned)
+            # add commas between brackets
+            cleaned = re.sub(r'\]\s+\[', '], [', cleaned)
+        # replace round brackets with square brackets
+        cleaned = cleaned.replace('(', '[').replace(')', ']')
```
michele-riva, line 43: Perhaps this should be done before the comma sanitization? Otherwise there would be no commas added between two round brackets if it is missing

src/viperleed/calc/files/new_displacements/tokens/linear_operation.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            msg = (
+                f'Invalid linear operation format: "{op_str.strip()}". '
+                'Expected a 1x1, 2x2, or 3x3 matrix.'
+            )
+            raise LinearOperationTokenParserError(msg)
+        self.arr = arr
+
+    @classmethod
+    def from_array(cls, arr) -> 'LinearOperationToken':
```
michele-riva, line 82:
```suggestion
    def from_array(cls, arr):
```

src/viperleed/calc/files/new_displacements/tokens/linear_operation.py
```diff
@@ -0,0 +1,96 @@
...
+        """Alternate constructor using numeric values directly."""
+        inst = cls.__new__(cls)
+        inst.arr = np.array(arr, dtype=float)
```
michele-riva, line 85: This implementation skips all the validation on arr compared to the string version. Perhaps the validation of the array should be factored out from init and reused here too?

src/viperleed/calc/files/new_displacements/tokens/offset.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, offset_str: str):
+        """Construct an OffsetToken from a string."""
+        parts = offset_str.strip().split()
+        if not (1 <= len(parts) <= 3):
+            msg = f'Invalid offset format (expect 1–3 floats): "{offset_str}".'
+            raise OffsetTokenParserError(msg)
+
+        try:
+            offset_values = np.array([float(p) for p in parts], dtype=float)
```
michele-riva, line 43: I think we can skip the conversion to a list of float here, and let numpy take care of it:

```suggestion
            offset_values = np.array(parts, dtype=float)
```

src/viperleed/calc/files/new_displacements/tokens/element.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Parameters
+    ----------
+    element_str : str
+        The string containing the element to be parsed.
+    """
+
+    def __init__(self, element_str: str):
```
michele-riva, line 36:
```suggestion
    def __init__(self, element_str):
```

src/viperleed/calc/files/new_displacements/tokens/offset.py
```diff
@@ -0,0 +1,82 @@
...
+    offset_str : str
+        The offset string to parse.
+
+    Raises
+    ------
+    OffsetTokenParserError
+        If parsing fails.
+    """
+
+    def __init__(self, offset_str: str):
```
michele-riva, line 35:
```suggestion
    def __init__(self, offset_str):
```

src/viperleed/calc/files/new_displacements/tokens/offset.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.offset = offset_values
+
+    @property
+    def dof(self):
+        """Return the number of degrees of freedom (DOF) for the offset."""
+        return self.offset.size
+
+    @classmethod
+    def from_floats(cls, *offsets: float) -> 'OffsetToken':
```
michele-riva, line 56:
```suggestion
    def from_floats(cls, *offsets):
```

src/viperleed/calc/files/new_displacements/tokens/offset.py
```diff
@@ -0,0 +1,82 @@
...
+        """Alternate constructor using 1–3 numeric values directly."""
+        if not (1 <= len(offsets) <= 3):
```
michele-riva, line 58: The validation code in this method is very similar to that in init. Consider refactoring to a _validate_offsets staticmethod

src/viperleed/calc/files/new_displacements/tokens/range_.py
```diff
@@ -0,0 +1,92 @@
...
+                raise RangeTokenParserError(msg) from err
+            setattr(self, attr, value)
+
+    @property
+    def has_step(self):
+        """Check if the range has a step defined."""
+        return self.step is not None
+
+    @classmethod
+    def from_floats(cls, start, stop, step=None) -> 'RangeToken':
```
michele-riva, line 59:
```suggestion
    def from_floats(cls, start, stop, step=None):
```

src/viperleed/calc/files/new_displacements/tokens/total_occupation.py
```diff
@@ -0,0 +1,64 @@
...
+    given as a single floating-point number that must be in the range
+    [0, 1].
+
+    Parameters
+    ----------
+    total_occupation_str : str
+        The string containing the total occupation to be parsed.
+    """
+
+    def __init__(self, total_occupation_str: str):
```
michele-riva, line 34:
```suggestion
    def __init__(self, total_occupation_str):
```

src/viperleed/calc/files/new_displacements/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Package new_displacements of viperleed.calc.files."""
+
+__authors__ = ('Alexandra Mia Imre (@alexmiame)',)
+__copyright__ = 'Copyright (c) 2019-2026 ViPErLEED developers'
+__created__ = '2024-10-04'
+__license__ = 'GPLv3+'
+
+from .file import DisplacementsFile
```
michele-riva, line 8: Same comment as for other package init files

src/viperleed/calc/files/new_displacements/file.py
```diff
@@ -0,0 +1,213 @@
...
+    provides a `next` method that takes the current R-factor as an argument
+    and returns the next search block.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.name = self._render_name
+        self._has_been_read = False
+        # an OFFSETS block is only allowed at the very beginning of the file
+        # we check this by setting this flag to True after the first block
```
michele-riva, line 55: Is this comment still relevant? _has_been_read is set to True as soon as we start reading, not after the first block. Perhaps this referred to an earlier implementation?

src/viperleed/calc/files/new_displacements/file.py
```diff
@@ -0,0 +1,213 @@
...
+                    for block in self.descendants
+                    if isinstance(block, OffsetsBlock)
+                ]
+            )
+            > 1
+        ):
+            raise OffsetsNotAtBeginningError(
+                'There can only be one OFFSETS block at the beginning of the '
+                'DISPLACEMENTS file.'
+            )
```
michele-riva, line 140: This bit seems redundant with the rest (though the comment is probably OK while the next one is misplaced)

src/viperleed/calc/files/new_displacements/segments.py
```diff
@@ -0,0 +1,367 @@
...
+    @abstractmethod
+    def is_my_header_line(line):
+        """Return whether `line` is a header for this segment."""
+
+    @property
+    def lines(self):
+        """Return the lines belonging to this segment."""
+        return tuple(self._lines)
+
+    def read_lines(self, lines):
```
michele-riva, line 69: Missing docstring

src/viperleed/calc/files/new_displacements/segments.py
```diff
@@ -0,0 +1,367 @@
...
+    def _render_name(self):
+        pass
+
+    @abstractmethod
+    def validate_segment(self):
+        """Validate the segment."""
+
+    @abstractmethod
+    def _belongs_to_me(self, line):
+        """Return whether `line` is a header for this segment."""
```
michele-riva, line 109: This docstring does not seem correct

src/viperleed/calc/files/new_displacements/segments.py
```diff
@@ -0,0 +1,367 @@
...
+
+    @staticmethod
+    def is_my_header_line(line):
+        """Loops are started by Loop start lines."""
+        return isinstance(line, LoopMarkerLine) and line.kind == 'start'
+
+    def _belongs_to_me(self, line):
+        """Loop end lines belong to this block."""
+        # is the loop already closed?
+        if self.lines and self.lines[-1].kind == 'end':
```
michele-riva, line 294: Probably should check that the last line is a LoopMarkerLine too, otherwise AttributeError on accessing .kind. Something like this would work:

```suggestion
        last_line = self.lines[-1] if self.lines else None
        if isinstance(last_line, LoopMarkerLine) and last_line.kind == 'end':
```

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,694 @@
...
+SECTION_HEADER_PATTERN = re.compile(
+    r'^=\s*(?P<section>GEO_DELTA|VIB_DELTA|OCC_DELTA|CONSTRAIN)$',
+    re.IGNORECASE,
+)
+OFFSETS_HEADER_PATTERN = re.compile(r'^=+\s*(?i:OFFSET[S]?)$', re.IGNORECASE)
+LOOP_START_PATTERN = re.compile(r'<loop>', re.IGNORECASE)
+LOOP_END_PATTERN = re.compile(r'<\\loop>|</loop>', re.IGNORECASE)
+
+
+_BELOW_DEBUG = 2
```
michele-riva, line 41: Better to replace this with LOG_VERY_VERBOSE from calc.constants

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,694 @@
...
+
+
+class OffsetsHeaderLine(HeaderLine):
+    """Class to parse the offsets header line in the DISPLACEMENTS file.
+
+    The offsets header line is of the form:
+        = OFFSETS
+    """
+
+    section = 'OFFSETS'
```
michele-riva, line 91: This class attribute seems unused in the implementation (e.g., in `__str__`)

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,694 @@
...
+
+class SectionHeaderLine(HeaderLine):
+    """Class to parse section header lines in the DISPLACEMENTS file.
+
+    Section header lines are of the form:
+        = <section>
+    where <section> is one of the defined sections in the DISPLACEMENTS file.
+    """
+
+    def __init__(self, section: str):
```
michele-riva, line 133:
```suggestion
    def __init__(self, section):
```

Remove also other type annotations below (and check also other modules)

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,694 @@
...
+            # It would be generally possible (e.g. link only z, but not xy for
+            # some atoms, reducing DOF from 6 to 5), but I'm not sure if one
+            # would realistically use this.
+
+        # parse targets
+        self.targets = self._parse_targets(targets_str)
+
+    def _treat_rhs(self):
+        # check for deprecated 'offset' tag
+        if 'offset' in self._rhs.lower():
```
michele-riva, line 501: I don't think that `offset` was ever supported in [CONSTRAIN blocks](https://www.viperleed.org/stable/content/calc/files/input/displacements/searchconstraints.html#searchconstraints). It (was) available as a left-hand-side flag for [GEO_DELTA](https://www.viperleed.org/stable/content/calc/files/input/displacements/geodelta.html#geodelta) and [VIB_DELTA](https://www.viperleed.org/stable/content/calc/files/input/displacements/vibdelta.html#vibdelta).

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,694 @@
...
+            if self.mode_token.mode is not PerturbationMode.OCC:
+                raise DisplacementsSyntaxError(
+                    'The "total" tag is only allowed for occupational '
+                    'constraints.'
+                )
+
+            # parse the total occupation value
+            total_occupation_str = self._rhs[len('total ') :].strip()
+
+            # this is a special case, where the linear operation is
```
michele-riva, line 530: comment seems incomplete

src/viperleed/calc/files/new_displacements/lines.py
```diff
@@ -0,0 +1,694 @@
...
+    def _format_rhs_str(self):
+        return f'{self.linear_operation} {self.link_target}'
+
+
+class OffsetsLine(ParsedLine):
+    """Class to parse lines in the OFFSETS block of DISPLACEMENTS.
+
+    Lines in the OFFSETS block are of the form:
+        <type> <target> [, <target> ...] [<direction>] = <offset>
+    where <target>, ... are tokens that are parsed by the OffsetToken class.
```
michele-riva, line 581: This does not seem correct

tests/calc/files/new_displacements/tokens/test_target.py
```diff
@@ -0,0 +1,133 @@
...
+    assert not regex.match('B')
+
+
+def test_invalid_number_format_raises():
+    with pytest.raises(TargetingError):
+        # non-integer in numeric list
+        TargetToken('A one two')
+    with pytest.raises(TargetingError):
+        # malformed layer spec
+        TargetToken('A L()')
```
michele-riva, line 123: Can be done in test_invalid_target_raises by adding these strings to the parametrization there

tests/calc/files/new_displacements/test_reader.py
```diff
@@ -0,0 +1,33 @@
...
+        # check correct type
+        assert isinstance(parsed, exp_cls)
+        # check correct attributes
+        for attr, value in exp_attrs.items():
+            assert getattr(parsed, attr) == value
+
+
+def test_displacements_reader(mock_displacements_path_and_lines):
+    path, expected_lines = mock_displacements_path_and_lines
+    _compare_lines(path, expected_lines)
```
michele-riva, line 26: One fails with `'ConstraintLine' object has no attribute 'type'`

tests/calc/files/new_displacements/test_segments.py
```diff
@@ -0,0 +1,106 @@
...
+    search_block = SearchBlock(header)
+    assert search_block.label == 'test'
+
+    returned_lines = search_block.read_lines(lines)
+    returned_lines = list(returned_lines)
+    # make sure all lines were processed
+    assert len(returned_lines) == 0
+
+
+def test_layered_loop_block():
```
michele-riva, line 48: Fails with `DisplacementsSyntaxError: Loop block cannot contain only another loop block.` at the read_lines call below

michele-riva (changes requested):
Some more suggestions on this one, but we're converging.

PR #429, opened 2025-06-11
doc: Spot Tracker doc v1.02

amimre:
Upload SpotTracker documentation v1.02.


amimre:
After discussion with @schmid-iap I also added a link to the release section in the viperleed-imagej repository where we will add previous versions of the Spot Tracker documentation.

PR #450, opened 2025-07-16
Search job class & move parabolaFit

amimre:
This PR refactors the subprocess handling in the TensErLEED search in `viperleed.calc`. This is necessary for compatibility with the `spawn()` based multiprocessing in #333.
Since the subprocess module can't use `spawn()` by itself, we now use multiprocessing to spawn a separate process and launch the subprocess from there.
For this, the subprocess handling has been moved to a dedicated `SearchJob` class.
This does not affect functionality, but hopefully also improves readability of the code.

While I was at it - as suggested by @michele-riva - I also move the parabolaFit functionality to a separate file, to clean search.py up a tiny bit more.


src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        self.command = command
+        self.input_data = input_data
+        self.log_path = Path(log_path) if log_path else None
+        self._proc = None
+        self._mp_proc = None
+        self._kill_me_flag = mp.Value("b", False)  # shared bool
+
+    @staticmethod
+    def _run_worker(command, input_data, log_path, kill_flag):
```
michele-riva, line 49: This could perhaps be a `@classmethod`, so you could use `cls._kill_proc_tree` rather than seemingly accessing a private attribute of a class.

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+        """Run the search command in a separate process."""
+
+        try:
+            log_f = open(log_path, "a") if log_path else subprocess.DEVNULL
```
michele-riva, line 53: Perhaps you could have a simple `@contextmanager` to handle this log file.

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+                pass
+            except (OSError, subprocess.SubprocessError):
+                logger.error(
+                    "Error starting search. Check files SD.TL and rf.info.")
+
+            while proc.poll() is None:
+                if kill_flag.value:
+                    SearchJob._kill_proc_tree(ps_proc)
+                    return
+                time.sleep(0.5)
```
michele-riva, line 84: May be nice to move this to a separate method too, so there is only one statement in this try.

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+            ps_proc.kill()
+        except psutil.NoSuchProcess:
+            pass
+
+    def start(self):
+        """Start the search subprocess inside a multiprocessing.Process."""
+        logger.debug('Starting search process with command '
+                     f'"{" ".join(self.command)}".')
+        self._mp_proc = mp.Process(
+            target=SearchJob._run_worker,
```
michele-riva, line 120:
```suggestion
            target=type(self)._run_worker,
```

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+            args=(self.command,
+                  self.input_data,
+                  str(self.log_path) if self.log_path else None,
```
michele-riva, line 123: Is there a reason to pass a string rather than a Path?

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+                  self._kill_me_flag)
+        )
+        self._mp_proc.start()
+
+    def is_running(self):
+        """Return True if the job is still running."""
+        return self._mp_proc.is_alive() if self._mp_proc else False
```
michele-riva, line 130:
```suggestion
        return self._mp_proc and self._mp_proc.is_alive()
```

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+        except KeyboardInterrupt:
+            logger.info("Killing process due to Keyboard interrupt.")
+            SearchJob._kill_proc_tree(ps_proc)
+            raise
+
+        except Exception as e:
+            try:
+                SearchJob._kill_proc_tree(ps_proc)
+            except Exception as ke:
+                raise
```
michele-riva, line 95:
```suggestion
            SearchJob._kill_proc_tree(ps_proc)
```
No need for a try if you just raise

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+                    SearchJob._kill_proc_tree(ps_proc)
+                    return
+                time.sleep(0.5)
+
+        except KeyboardInterrupt:
+            logger.info("Killing process due to Keyboard interrupt.")
+            SearchJob._kill_proc_tree(ps_proc)
+            raise
+
+        except Exception as e:
```
michele-riva, line 91: `e` unused

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,146 @@
...
+import subprocess
+import time
+
+logger = logging.getLogger(__name__)
+
+class SearchJob:
+    """Launch and manage a the MPI-based search job in a separate
+    multiprocessing.Process, taking care of directing the input to stdin, and
+    logging the output. Handles graceful termination of all MPI-subprocesses
+    KeyBoardInterrupt and monitoring.
```
michele-riva, line 23: First docstring line should be a full sentence.

michele-riva (changes requested):
Looks good. A few suggestions like always.

michele-riva:
Sorry, forgot one thing in the review: it would be great to have tests for the new class

amimre:
I implemented some of the suggestions.
Tests here are a bit tricky since the class essentially just manages a subprocess and monitors/kills it as necessary. All of that is hard to test without extensive mocking – at which point the tests become kind of useless anyhow...

michele-riva (changes requested):
Some more small things to be done, sorry

amimre:
Concerning the formatting: I'd like to make another appeal for implementing #276. Taking care of formatting manually (and then manually checking it in review) is a waste of time on everyone's part.

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,169 @@
...
+        """Start the search subprocess inside a multiprocessing.Process."""
+        logger.debug(
+            'Starting search process with command ' f'"{" ".join(self.command)}".'
+        )
+        self._mp_proc = mp.Process(
+            target=_run_search_worker,
+            args=(
+                self.command,
+                self.input_data,
+                str(self.log_path),
```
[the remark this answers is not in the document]
amimre: As mentioned, it's better to use the string in this case because simple objects are easier to transfer across process boundaries.
I'll resolve and use None.

tests/calc/classes/test_search_job.py
```diff
@@ -0,0 +1,39 @@
+"""Class SearchJob."""
+
+__authors__ = ("Alexander M. Imre (@amimre)",)
```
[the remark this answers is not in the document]
amimre: Unnecessary if there is only one name. Auto-formatting trims it to one line.

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,169 @@
...
+        f = open(log_path, "a")
+        try:
+            yield f
+        finally:
+            f.close()
+    else:
+        yield subprocess.DEVNULL
+
+
+def _run_search_worker(command, input_data, log_path, kill_flag, return_code):
```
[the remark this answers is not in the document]
amimre: I intentionally didn't make it into another class because it's clearer to check what is happening on either side of the "multiprocessing boundary". One has to be quite careful there to not mess this up, and the single _run_search_worker method seemed to me to be the clearest option.

michele-riva (changes requested):
Sorry, @amimre, a few more suggestions, and one test to cross-check.

tests/calc/classes/test_search_job.py
```diff
@@ -0,0 +1,99 @@
...
+        assert job.is_running()
+    job._kill_me_flag.value = True
+    job.wait()  # ensure cleanup
+
+    with subtests.test('is not running after kill'):
+        assert not job.is_running()
+
+
+@pytest.mark.timeout(5)
+def test_notice_immediate_job_death(capfd):
```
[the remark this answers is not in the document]
amimre: It does. Can you try to increase the timeout to see if that fixes it?
Otherwise, it may be a platform issue.

src/viperleed/calc/classes/search_job.py
```diff
@@ -0,0 +1,168 @@
...
+        logger.debug(f'Starting search process with command {command_str}.')
+        self._mp_proc = mp.Process(
+            target=_run_search_worker,
+            args=(
+                self.command,
+                self.input_data,
+                str(self.log_path.resolve()) if self.log_path else None,
+                self._kill_me_flag,
+                self._return_code,
+            ),
```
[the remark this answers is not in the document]
michele-riva: Too bad. We should find a different auto-formatter then. https://github.com/google/yapf may be an option.

tests/calc/classes/test_search_job.py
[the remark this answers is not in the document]
michele-riva: Unfortunately still failing. `CaptureResult.err == ''` on my machine.

PR #454, opened 2025-07-30
Fix bad pixel finding RAM use

FlorianDoerr:
Reimplement frame saving to reduce RAM use.


src/viperleed/gui/measure/camera/badpixels.py
```diff
@@ -1415,3 +1405,34 @@ def __has_info(self):
     def __bool__(self):
         """Return the truth value of self."""
         return bool(self.__has_info and self.n_bad_pixels_sensor)
+
+
+class BadPixelsFrameStorage():
```
michele-riva, line 1410:
```suggestion
class BadPixelsFrameStorage:
```

michele-riva (changes requested):
Good job @FlorianDoerr. Only very small suggestions.

src/viperleed/gui/measure/camera/badpixels.py
```diff
@@ -1415,3 +1406,33 @@ def __has_info(self):
...
+        self._sum += frame
+        self._sum_squares += frame**2
+
+    def mean(self):
+        """Return the mean of the stored frames."""
+        return self._sum / self._count
+
+    def var(self):
+        """Return the variance calculated from the stored frames."""
+        return (self._sum_squares - self._sum**2 / self._count) / self._count
```
michele-riva, line 1438:
```suggestion
        mean_of_squares = self._sum_squares / self._count
        return mean_of_squares - self.mean()**2
```

For reading it more easily (unless there may be some uncaught overflow errors).

michele-riva (changes requested):
Very nice. Two tiny suggestions left.

michele-riva (dismissed):
Good to go!

PR #457, opened 2025-08-12
Always re-normalize weights to sum to 1 when averaging beams

### LINKED ISSUE #453: Domain calculation - Intensities of superpos file too large
In domain calculations the final superpos run creates a superpos-spec.out file, where all intensities are generally too large by a factor of exactly 100. This happens in any domain calculation, already in the tleedm.09 version and is still the case in viperleed 12.2. It seems that the percentage values of the domain weights were falsely taken as integer numbers. 
This bug doesn`t produce any wrong result as long as all intensities remain below 0.01. That's presumably why it wasn't detected so far. If, however, the intensity of one single beam and energy exceeds this value (as frequently e.g. for the (0 0) beams), then the R-factor program  crashes with error message:
MAX INTENSITY IN THEOR BEAM   1 IS SUSPECT(  0.10511E+01)- **** STOP PROGRAM ****

fkraushofer:
Fixes #453

As @LutzHammer already suspected, the issue is caused by the "percentages" in SD.TL being taken as-is, without dividing by 100, here:
https://github.com/viperleed/viperleed/blob/2cefa8dac9fe16b0ef24cf4ec830c3e45dcfb468/src/viperleed/calc/sections/superpos.py#L196-L213

Rather than fixing it directly there, this PR edits `viperleed.calc.files.beams.averageBeams` to always re-normalize weights such that they sum to 1 before averaging. This should solve this type of issue more generally.

### DIFF
--- a/src/viperleed/calc/files/beams.py
+++ b/src/viperleed/calc/files/beams.py
@@ -33,11 +33,14 @@ logger = logging.getLogger(__name__)
 
 def averageBeams(beams, weights=None):
     """Takes a list of percentages and a list of lists of Beam objects.
-    Returns a new list of Beam objects with weighted averaged intensities."""
+    Returns a new list of Beam objects with weighted averaged intensities.
+    Weights will be renormalized such that they sum to 1."""
     if beams is None:
         raise ValueError("averageBeams: No beams passed.")
     if weights is None or len(weights) == 0:
-        weights = [1/len(beams)] * len(beams)
+        weights = np.repeat([1/len(beams)], len(beams))
+    else:
+        weights = np.array(weights) / sum(weights)
     avbeams = copy.deepcopy(beams[0])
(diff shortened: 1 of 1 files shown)


michele-riva (changes requested):
Looks good. Perhaps it would be nice to have a couple simple tests to check the correct normalization of the weights (both when given and when not given as a kwarg).

fkraushofer:
Still made one minor change.

I agree that tests would be nice, though more for the rest of the function than for the weights (I don't really need unit tests to diagnose that this one-line renormalization will work correctly). However, I don't have time to do that properly right now and neither do you, so I would say let's merge this as a quick bugfix for now, and adding tests will come up again anyway as part of the bigger cleaning-up-domains milestone.

michele-riva (approved):

PR #458, opened 2025-09-02
Global configuration path

FlorianDoerr:
- Creates a new Measurement.ini file that replaces the old _system_settings.ini in its function at a path generated by QSettings.
- Copies old _defaults to new path genereated by QSettings aswell.


michele-riva:
Addendum to the above: Do you think it would make sense to also modify the firmware uploader in this PR so that it copies its settings file to the new location? Or should we leave this for a separate PR?

src/viperleed/gui/base.py
```diff
@@ -17,6 +17,7 @@
 import sys
 
 import numpy as np
+from PyQt5 import QtCore as qtc
```
michele-riva, line 20: Actually, it would be better to move this to its own module. Stuff in `gui.base` is imported also in `calc`, which does not have `PyQt5` as a mandatory dependency currently. Perhaps create a `gui/qsettings.py` module.

src/viperleed/gui/base.py
```diff
@@ -701,3 +702,20 @@ def orientation(vector, zero_pi=True, precision=4):
     if not zero_pi:
         return angle
     return angle % 180
+
+
+def get_qsettings(settings):
```
michele-riva, line 707: Perhaps `settings`->`file_name`. That way it is clear that this will search (and create) a file with that name.

src/viperleed/gui/measure/hardwarebase.py
```diff
@@ -247,6 +242,13 @@ def disconnected_slot(slot, signal, *more_signals, type=None):
             safe_connect(signal, slot, type=type)
 
 
+def get_default_path():
+    """Return the path to default settings."""
+    # The path is detected at runtime and not set as a global variable
+    # to catch any changes that may occur to the path in runtime.
+    return Path(get_qsettings('Measurement').fileName()).parent.resolve()
```
michele-riva, line 249:
```suggestion
    return Path(get_qsettings('Measurement').fileName()).resolve().parent
```

Better to `.resolve()` before `.parent`

michele-riva (changes requested):
This is also good, except for my note about the `gui.base` module and its interaction with `calc`.

michele-riva (approved):

PR #459, opened 2025-09-11
Aliases and future backwards compatibility

### LINKED ISSUE #242: ViPErLEEDSettings backwards compatible value getting
We should reimplement getting values from a settings file in a manner that allows checking for aliases to simplify ensuring backwards compatibility in the future. We could achieve this by having an internal file that keeps a list of aliases for the name of the desired value and after getting that value failed we would try to get each name in that list.

We should rename measurement_devices from the controller section of the controller settings to measurable_quantities.

FlorianDoerr:
Backwards compatible value getting with aliases.

Fixes #242


src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -205,6 +205,7 @@
 from viperleed.gui.measure.classes.decorators import emit_default_faulty
 from viperleed.gui.measure.classes.datapoints import DataPoints
 from viperleed.gui.measure.classes.settings import DefaultSettingsError
+from viperleed.gui.measure.classes.settings import ensure_aliases_exist
```
michele-riva, line 208: small caps come alphabetically (in ASCII) after capital letters, so this import should go after all the classes.

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -36,6 +38,21 @@
 from viperleed.gui.measure.widgets.pathselector import PathSelector
 
 
+def ensure_aliases_exist():
+    """Merge default and user aliases and create file if required."""
+    _tmp = ConfigParser()
+    _tmp.read(Path(__file__).parent.parent / '_defaults' / '_aliases.ini')
+    _tmp.read(get_aliases_path())
+    with open(get_aliases_path(), 'w', encoding='utf-8') as fproxy:
+        _tmp.write(fproxy)
```
michele-riva, line 47:
```suggestion
    user_aliases = get_aliases_path()
    _tmp.read(user_aliases)
    with user_aliases.open('w', encoding='utf-8') as fproxy:
        _tmp.write(fproxy)
```

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -94,7 +111,200 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+                    value = super().get(sec, opt, raw=raw, vars=vars)
+                except (NoSectionError, NoOptionError):
+                    continue
+                else:
+                    self[section][option] = value
+                    self.remove_option(sec, opt)
+                    break
+            else:
+                value = fallback
+            if value == _UNSET:
```
michele-riva, line 184:
```suggestion
            if value is _UNSET:
```

(I think `_UNSET` is a singleton.)

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -94,7 +111,200 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+        ------
+        NoSectionError
+            If the settings section is missing.
+        NoOptionError
+            If the settings option is missing.
+        """
+        value = _UNSET
+        try:
+            value = super().get(section, option, raw=raw, vars=vars)
+        except (NoSectionError, NoOptionError) as err:
```
michele-riva, line 172:
```suggestion
        except (NoSectionError, NoOptionError):
```

Since you don't use `err` for anything other than re-raising, you can skip the "as" part here, and just use a bare `raise` instead of `raise err`.

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -94,7 +111,200 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+                try:
+                    value = super().get(sec, opt, raw=raw, vars=vars)
+                except (NoSectionError, NoOptionError):
+                    continue
+                else:
+                    self[section][option] = value
+                    self.remove_option(sec, opt)
+                    break
+            else:
+                value = fallback
```
michele-riva, line 183: Perhaps clearer if this would be its own private method? E.g., `_get_from_alias(section, option, raw=raw, vars=vars, fallback=_UNSET)`. It could then be made a little simpler by replacing `break` with `return value`, and replacing the `else` block with `return fallback`.

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -94,7 +111,200 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+            if value == _UNSET:
+                raise err
+
+        if value == '':
```
michele-riva, line 187:
```suggestion
        if not value:
```

(This would also catch `None`, not sure if it is desirable.)

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -94,7 +111,200 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+        option : str
+            The option key.
+
+        Returns
+        -------
+        aliases : list
+            A list of section/option pairs.
+        """
+        try:
+            aliases = self._aliases[section + '/' + option]
```
michele-riva, line 250:
```suggestion
            aliases = self._aliases[f'{section}/{option}']
```

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -94,7 +111,200 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+            relevant = conv(aliases.get(self._cls_name, 'relevant_sections'))
+        except NoSectionError:
+            # The settings for this class do not have any aliases.
+            return
+        except NoOptionError:
+            # No other relevant sections.
+            relevant = []
+
+        for section in (self._cls_name, *relevant):
+            for key in aliases[section].keys():
```
michele-riva, line 279:
```suggestion
            for key in aliases[section]:
```

Probably enough to iterate without `.keys()`?

src/viperleed/gui/measure/_defaults/_aliases.ini
```diff
@@ -0,0 +1,29 @@
...
+[ViPErinoController]
+relevant_sections = ('ControllerABC', )
+
+[MeasurementABC]
+new_sections = ('energies', )
+fallback_values = (('energies/step_profile', 'abrupt'),)
+energies/start_energy = ('measurement_settings/start_energy', )
+energies/delta_energy = ('measurement_settings/delta_energy', )
+energies/end_energy = ('measurement_settings/end_energy', )
+energies/step_profile = ('measurement_settings/step_profile', )
```
michele-riva, line 15: Would it make sense to just have an alias for the whole section rather than listing all its contents?

src/viperleed/gui/measure/_defaults/_aliases.ini
```diff
@@ -0,0 +1,29 @@
+[ControllerABC]
+measurement_settings/nr_samples = ('measurement_settings/num_meas_to_average', )
+controller/serial_class = ('controller/serial_port_class', )
+controller/address = ('controller/port_name', )
+
+[ViPErinoController]
+relevant_sections = ('ControllerABC', )
```
michele-riva, line 7: Perhaps let's discuss this name. It may be confused with "new_sections".

michele-riva (changes requested):
Very nice work @FlorianDoerr. Only very small suggestions.

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -92,7 +109,228 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+            if value is _UNSET:
+                raise
+
+        if value == '':
+            try:
+                value = self._fallback[section][option]
+            except KeyError:
+                pass
+            else:
+                self[section][option] = value
```
michele-riva, line 182: Perhaps add a comment here: the fallback that you're storing is not the one given by the user, but the one that is stored in the aliases.

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -92,7 +109,228 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+        -------
+        None.
+        """
+        self._aliases = {}
+        self._fallback = defaultdict(dict)
+        self._cls_name = cls_name
+        if not self._cls_name:
+            # No class name, therefore we cannot look up aliases.
+            return
+        self._prepare()
```
michele-riva, line 204: Maybe `_load_aliases()`?

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -92,7 +109,228 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+            try:
+                value = super().get(sec, opt, raw=raw, vars=vars)
+            except (NoSectionError, NoOptionError):
+                continue
+            else:
+                self[section][option] = value
+                self.remove_option(sec, opt)
+                return value
+        if section == 'energies':
+            raise RuntimeError('How did I get here.')
```
michele-riva, line 278: Add comment for this one, or remove it if it was a temporary check

michele-riva (changes requested):
A few very small suggestions left on this one.

src/viperleed/gui/measure/_defaults/_aliases.ini
```diff
@@ -0,0 +1,29 @@
+[ControllerABC]
```
michele-riva, line 1: Perhaps it would be good to add a comment at the beginning of this file that will help us in the future to add more aliases. Things to mention: (1) describe section names; (2) structure of keys and values for aliases (i.e., the key is new_section/new_option, values are tuples of 'old_section/old_option' strings); (3) describe special keys (i.e., new_sections, parent_aliases, fallback_values).

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -92,7 +109,230 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+        Returns
+        -------
+        aliases : list
+            A list of section/option pairs.
+        """
+        try:
+            aliases = self._aliases[f'{section}/{option}']
+        except KeyError:
+            return []
+        return [alias.split('/') for alias in aliases]
```
michele-riva, line 244:
```suggestion
    def _iter_aliases(self, section, option):
        """Yield aliases for a section/option pair.

        Parameters
        ----------
        section : str
            The section key.
        option : str
            The option key.

        Yields
        ------
        alias : tuple
            Names of old section/option corresponding to the ones given.
        """
        try:
            aliases = self._aliases[f'{section}/{option}']
        except KeyError:
            return
        yield from (alias.split('/') for alias in aliases)
```

Since we only ever iterate over this (and break off at the first success) it may be worth making this an iterator. This prevents creation of all the section/option tuples every time.

tests/gui/measure/classes/test_settings.py
```diff
@@ -0,0 +1,298 @@
...
+''')
+        mocker.patch(f'{_MODULE}.get_aliases_path', return_value=aliases)
+
+    def test_fallback_from_aliases(self):
+        """Check retrieval of a fallback_value from the aliases."""
+        parser = AliasConfigParser(cls_name='WithAliases')
+        parser.read_dict({'A': {'opt': ''}})
+        expect = 'fb'
+        assert parser.get('A', 'opt') == expect
+        # After .get, the value is permanently stored
```
michele-riva, line 196: These comments from me are not relevant anymore now that we replace the aliases right away. I think the code can stay though.

michele-riva (changes requested):
Some suggestions while you're still changing the code. See what still applies. Feel free to get resolve the rest.

tests/gui/measure/_defaults/test_aliases.py
```diff
@@ -0,0 +1,92 @@
...
+import ast
+from configparser import ConfigParser
+
+from pytest_cases import fixture
+
+from viperleed.gui.measure.classes.settings import get_default_aliases_path
+
+
+@fixture
+def alias_config():
```
michele-riva, line 20: If you run pylint on this file, it will tell you that your test functions "mask" the "global" alias_config name. A way to prevent that is the following:
```suggestion
@fixture(name='alias_config')
def fixture_alias_config():
```

tests/gui/measure/_defaults/test_aliases.py
```diff
@@ -0,0 +1,92 @@
...
+            # Check if any of the parent_aliases are among
+            # the sections that declare parent_aliases.
+            for parent in parents:
+                assert parent not in inheriting_sections, (
+                    f'Invalid multiple inheritance: Section "{section}"'
+                    f' tries to inherit from "{parent}", which already '
+                    'has its own parent_aliases.'
+                    )
+
+    def test_no_capital_letters_in_options_and_values(self, alias_config):
```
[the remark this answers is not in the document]
michele-riva: Perhaps we should extend this no-capitalized-sections-and-options test also to all other files in the _defaults (skipping `_system_settings.ini` that, unfortunately, historically has `[PATHS]`).

tests/gui/measure/_defaults/test_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+                    assert value == value.lower(), (
+                        f'Alias "{value}" in [{section}][{option}] '
+                        'contains uppercase letters.'
+                        )
+                    continue
+                # Treat key 'fallback_values' differently because the
+                # fallback values may contain capitalized letters.
+                for value in ast.literal_eval(alias_config[section][option]):
+                    assert value[0] == value[0].lower(), (
+                        f'Alias "{value[0]}" in [{section}][{option}] '
```
michele-riva, line 57:
```suggestion
                        f'Fallback "{value[0]}" in [{section}][{option}] '
```

tests/gui/measure/_defaults/test_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+                )
+
+    def test_no_multiple_inheritance(self, alias_config):
+        """Check that there is no multiple inheritance."""
+        # Collect all sections that declare parent_aliases themselves.
+        inheriting_sections = (
+            section
+            for section in alias_config.sections()
+            if alias_config.has_option(section, 'parent_aliases')
+            )
```
michele-riva, line 90: Watch out, this is a generator that will be exhausted at the first iteration. Use a set instead:
```suggestion
        inheriting_sections = {
            section
            for section in alias_config.sections()
            if alias_config.has_option(section, 'parent_aliases')
            }
```

tests/gui/measure/classes/test_settings.py
```diff
@@ -0,0 +1,270 @@
...
+        user_aliases.write_text('[Foo]\nchanged=user')
+        installed_aliases = ConfigParser()
+        installed_aliases.read_string('[Foo]\nstays=stays\nchanged=installed')
+        defaults = tmp_path / '_defaults'
+        defaults.mkdir()
+        with (defaults/'_aliases.ini').open('w') as installed_ini:
+            installed_aliases.write(installed_ini)
+        mocker.patch(f'{_MODULE}.get_aliases_path', return_value=user_aliases)
+        mocker.patch(f'{_MODULE}.SRC_ALIASES_PATH',
+                     str(tmp_path / '_defaults' / '_aliases.ini'))
```
michele-riva, line 58:
```suggestion
        mocker.patch(f'{_MODULE}.SRC_ALIASES_PATH', defaults / '_aliases.ini')
```

tests/gui/measure/classes/test_settings.py
```diff
@@ -0,0 +1,270 @@
...
+        merged_aliases = ConfigParser()
+        merged_aliases.read(user_aliases)
+        assert merged_aliases['Foo']['stays'] == 'stays'
+        assert merged_aliases['Foo']['changed'] == 'user'
+
+    def test_new_aliases_written(self, tmp_path, mocker):
+        """Check that a user aliases.ini file is always created."""
+        user_aliases = tmp_path / 'aliases.ini'
+        mocker.patch(f'{_MODULE}.get_aliases_path', return_value=user_aliases)
+        mocker.patch(f'{_MODULE}.__file__', str(tmp_path / 'dummy/dummy.py'))
```
michele-riva, line 69: This line is not needed anymore. You could patch `SRC_ALIASES_PATH` if you want, but the test will work anyway.

tests/gui/measure/classes/test_settings.py
```diff
@@ -0,0 +1,270 @@
...
+        """Check complaints for a missing non-aliased option."""
+        parser = AliasConfigParser(cls_name='SomeClass')
+        parser.read_dict({'Foo': {}})
+        with pytest.raises(NoOptionError):
+            parser.get('Foo', 'opt')
+
+    def test_missing_section_raises(self):
+        """Check complaints for a missing non-aliased section."""
+        parser = AliasConfigParser(cls_name='SomeClass')
+        parser.read_dict({'Foo':{}})
```
michele-riva, line 160:
```suggestion
        parser.read_dict({'Foo': {}})
```

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -92,7 +110,156 @@ class TooManyDefaultSettingsError(DefaultSettingsError):
...
+        except NoOptionError:
+            # No other relevant sections.
+            relevant = []
+
+        for section in (self._cls_name, *relevant):
+            for key in aliases[section]:
+                if key == 'new_sections':
+                    continue
+                if key == 'fallback_values':
+                    self._fallback = conv(aliases[section][key])
```
michele-riva, line 209: What if both child and parent define some fallbacks (for potentially different targets)?

michele-riva (changes requested):
Some more suggestions concerning the fallbacks. I'd also suggest to expand the tests a tad for checking the handling of fallbacks. Also, I did not see yet the tests for the other .ini files. I'd suggest placing them into tests/gui/measure/_defaults/test_other_defaults.py (I though of calling it test_device_defaults, but there are also those of measurements).

PR #466, opened 2025-10-09
Fix delta-amplitude calculation in domains

### LINKED ISSUE #440: Domain calculation crashes during section "DELTA-AMPLITUDES"
As of version 0.13.2, viperleed calc crashes during the delta calculations when I run a domain calculation (RUN=4). With viperleed version 0.12.2 the calculation of the delta amplitudes terminates successfully with the same input files. The complete run (RUN=4) is possible without crashing. So, the problem does not originate from the directory tree structure or from an error in the input files (e.g. DISPLACEMENTS). At least I didn't find any errors.

Attached you find the viperleed-calc.log file with the debug output, the DISPLACEMENTS, and PARAMETERS file. Let me know, if you need more files.

[viperleed-calc-250623-144205.log](https://github.com/user-attachments/files/20865808/viperleed-calc-250623-144205.log)
[DISPLACEMENTS.txt](https://github.com/user-attachments/files/20865977/DISPLACEMENTS.txt)
[PARAMETERS.txt](https://github.com/user-attachments/files/20865978/PARAMETERS.txt)

michele-riva:
Fixes #440.

Also takes care of a few other bugs:
- prevent double writing of modified `PARAMETERS` with identical value. Important for `DOMAIN`, as only one path can be specified for a given domain.
- prevent `AttributeError` in `iosearch` (renamed attributes missed in earlier commit)
- prevent useless rerunning of the reference calculation for domains that already have a consistent Tensor: write to domain `PARAMETERS` important values inherited from the main one, duplicate `IVBEAMS` from main.
- ensure domains use the same `TL_VERSION` as specified in the main `PARAMETERS`.
- ensure reference calculations are executed for all domains after a search when executed in the same calc run.
- ensure a reference calculation is executed for all domains (even if not needed) if the user explicitly requested one.
- prevent duplicate items within `Tensors.zip` and `Deltas.zip` archives

This PR also collects changes from:
- #449
- #441 

Remaining TODOs:
- [x] ensure all tests run

### DIFF
--- a/src/viperleed/calc/sections/search.py
+++ b/src/viperleed/calc/sections/search.py
@@ -521,7 +521,10 @@ def search(sl, rp):
         libpattern = 'lib.search'
         if usempi and rp.TL_VERSION <= Version('1.7.3'):
             libpattern += '.mpi'
-        lib_file = next(libpath.glob(libpattern + '*'), None)
+        lib_files = libpath.glob(libpattern + '*')
+        if not usempi:
+            lib_files = (f for f in lib_files if 'mpi' not in f.name)
+        lib_file = next(lib_files, None)
         if lib_file is None:
             raise FileNotFoundError(f'File {libpattern}.f not found.')
 


tests/_test_data/domains/telluride_stacking/deltas/PARAMETERS
```diff
@@ -0,0 +1,10 @@
+! ####### GLOBAL PARAMETERS #######
+RUN = delta
```
fkraushofer: For a test, this is maybe not the best; would crash if Tensors are not exactly in the expected format. I think a better test would be `RUN = domains` (also checks whether refcalc is needed or not) or even `RUN = 1-3` (to enforce a reference calculation).

fkraushofer (changes requested):
Sorry for the delay. Changes look good to me in general, one minor comment below). I started some tests now (one "standard" single-domain calculation with a simple system, the other one with the input from #440), these are waiting in the queue, I will report (+ approve) once they are done. For the domain test, I now used `RUN = 4`, which I think is the better test (see also below).

fkraushofer:
Single-domain test ran without errors.
Test with the input from #440 failed. With `RUN = 4`, `viperleed calc` correctly identified that the tensors are compatible, but then in the domain calculation did not find Tensors files with the expected names. I suspect this is because the tensor name conventions changed. I'll re-try with `RUN = 1-3`. However, there's a minor issue/bug apparent in the log that maybe @michele-riva could already have a look at:

```
STARTING SECTION: INITIALIZATION (DOMAINS)
dbg: IVBEAMS file was read successfully
Running with backend TensErLEED.
Fetching input files for domain STA
Unpacking Tensors_001.zip...
Reading input files for domain STA
dbg: Skipping line 1 in input file: '! ####### PARAMETERS FOR INTERPRETING THE POSCAR #######'.
dbg: Skipping line 2 in input file: ''.
dbg: Skipping line 6 in input file: ''.
dbg: Skipping line 8 in input file: ''.
dbg: Skipping line 9 in input file: '! ####### PARAMETERS FOR REFCALC #######'.
dbg: Skipping line 10 in input file: ''.
dbg: Skipping line 12 in input file: ''.
dbg: Skipping line 13 in input file: ''.
dbg: Skipping line 14 in input file: '! ######################################################'.
dbg: Skipping line 15 in input file: '! #  THE FOLLOWING LINES WERE GENERATED AUTOMATICALLY  #'.
dbg: Skipping line 16 in input file: '! ######################################################'.
dbg: Skipping line 17 in input file: ''.
```

Here's the full log: [viperleed-calc-251017-225757.log](https://github.com/user-attachments/files/22989505/viperleed-calc-251017-225757.log)

fkraushofer:
Third attempt also failed. Same behavior as first attempt, log reports this despite `RUN = 1-3`:
```
STARTING SECTION: REFERENCE CALCULATION (DOMAINS)
Found no domain which requires a reference calculation.
Finishing section at 11:40:13. Section took 0.00 seconds.
```

I would consider that a bug. If a reference calculation is explicitly asked for in `RUN`, it should be executed for all domains, same as in a single-domain run.

michele-riva:
Thanks @fkraushofer for looking into this. OK, there seems to be a bit of a mess when running from the original inputs in #440. I admit I have tried running from a "curated" version of those inputs, but I will now try the full stuff and dissect the problems. I will let you know when I have something stable to test.

michele-riva:
> However, there's a minor issue/bug apparent in the log that maybe @michele-riva could already have a look at:
> 
> ```
> STARTING SECTION: INITIALIZATION (DOMAINS)
> dbg: IVBEAMS file was read successfully
> Running with backend TensErLEED.
> Fetching input files for domain STA
> Unpacking Tensors_001.zip...
> Reading input files for domain STA
> dbg: Skipping line 1 in input file: '! ####### PARAMETERS FOR INTERPRETING THE POSCAR #######'.
> dbg: Skipping line 2 in input file: ''.
> dbg: Skipping line 6 in input file: ''.
> dbg: Skipping line 8 in input file: ''.
> dbg: Skipping line 9 in input file: '! ####### PARAMETERS FOR REFCALC #######'.
> dbg: Skipping line 10 in input file: ''.
> dbg: Skipping line 12 in input file: ''.
> dbg: Skipping line 13 in input file: ''.
> dbg: Skipping line 14 in input file: '! ######################################################'.
> dbg: Skipping line 15 in input file: '! #  THE FOLLOWING LINES WERE GENERATED AUTOMATICALLY  #'.
> dbg: Skipping line 16 in input file: '! ######################################################'.
> dbg: Skipping line 17 in input file: ''.
> ```
> 
> Here's the full log: [viperleed-calc-251017-225757.log](https://github.com/user-attachments/files/22989505/viperleed-calc-251017-225757.log)

About this: do you mean the various debug records about skipping lines? I've never been a huge fan of those either, as they clutter the log quite a bit. I can suppress them, or perhaps give them an even lower level than debug. Then they would still be accessible for dev debugging purposes should anything go wrong there.

src/viperleed/calc/sections/deltas.py
```diff
@@ -57,19 +59,23 @@ class DeltaCompileTask:
     source_dir : Path
         Path to the folder containing the static Fortran
         source files to be compiled.
+
+    Notes
+    -----
+    It is important to create instances of this class while the current
+    directory is the work directory for the specific domain: the main
+    work directory for a single-domain calculation, the relevant
+    subfolder of the main work directory for multi-domain ones.
```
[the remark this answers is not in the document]
fkraushofer: Much better, thanks

amimre:
If possible, I would maybe keep these messages at the lowest log level since it can be useful for some parser debugging. (`LOG_DEBUG=1` I think, which should be equivalent to running with `-vvv`).

michele-riva:
About the skipped lines: I have now moved them to the `LOG_VERY_VERBOSE` level as suggested by @amimre. The primary reason is that the behavior comes from the `InputFileReader` class, so it is common to all file readers based on it (currently PARAMETERS and the new DISPLACEMENTS parser, but will eventually apply to all input files). I guess it is however safe to not print empty skipped lines, as we are completely agnostic to those. @amimre: any objection?

michele-riva:
> Third attempt also failed. Same behavior as first attempt, log reports this despite `RUN = 1-3`:
> 
> ```
> STARTING SECTION: REFERENCE CALCULATION (DOMAINS)
> Found no domain which requires a reference calculation.
> Finishing section at 11:40:13. Section took 0.00 seconds.
> ```
> 
> I would consider that a bug. If a reference calculation is explicitly asked for in `RUN`, it should be executed for all domains, same as in a single-domain run.

This is the result of the fix for the contents of `Tensors.zip` files that we have discussed. By the time we enter the refcalc, there is currently no recalling of whether users had set `RUN=4` or `RUN=1-3`, since we replace occurrences of `4` in `RUN` with the latter (or with `2-3` if no domain needs a refcalc). This is in fact inconsistent with what we say in the documentation (see end of paragraph after [this figure](https://www.viperleed.org/stable/content/calc/sections/domain_calculation.html#id3)). In order to force a refcalc for all domains we would have to explicitly set `refcalc_required` on all of them before we do the replacement. Does that sound reasonable?

fkraushofer:
> This is the result of the fix for the contents of `Tensors.zip` files that we have discussed. By the time we enter the refcalc, there is currently no recalling of whether users had set `RUN=4` or `RUN=1-3`, since we replace occurrences of `4` in `RUN` with the latter (or with `2-3` if no domain needs a refcalc). This is in fact inconsistent with what we say in the documentation (see end of paragraph after [this figure](https://www.viperleed.org/stable/content/calc/sections/domain_calculation.html#id3)). In order to force a refcalc for all domains we would have to explicitly set `refcalc_required` on all of them before we do the replacement. Does that sound reasonable?

I feel like this only solves the inverse problem, unfortunately: It would allow us to always force a refcalc if the user had specified `RUN = 4` or something like `RUN = 1 4`. But the more problematic case is actually the one where the user specifies `RUN = 1-3` - the expected behavior would be to run a refcalc regardless of whether one is required or not - otherwise `RUN` behaves completely differently between domain and non-domain calculations. In most real-world usecases, I don't really see myself using `RUN = 4`, but instead treating the domains similar to every other calculation, where I want direct control over whether a new refcalc should be done or not. Normally, whenever `RUN = 1` is set, we do a reference calculation, no questions asked. If it does an under-the-hood check, as in these tests, then there is no obvious way to force a reference calculation if `calc` (incorrectly) assumes none is needed.

I would suggest that instead of replacing all instances of `4` in `RUN` during intialization, we instead treat them when we get to them during the loop, right here:
https://github.com/viperleed/viperleed/blob/89d05aa2ac85672e74623898d9e0c91d728d188d/src/viperleed/calc/sections/run_sections.py#L263-L265

Code mockup to insert could be:
```
if sec == 4:
    # go to refcalc segment now, decide automatically whether they are required for domains.
    sec = 1
    # Then do deltas and search.
    rp.RUN = [2, 3] + rp.RUN
elif sec == 1:
    # if this is a domain calculation, set refcalc_required = True for all domains
    # this will *not* trigger for RUN = 4
    ...
```

fkraushofer:
Oh, that's true! I mistook what https://github.com/viperleed/viperleed/commit/5eef44e89066d36b70b0f81bffefb39f85ffd409 would end up doing, but you're right, that should do the trick. Philosophically where to put it - I think both is fine. Yes, most index-dispatching happens in the section loop, but on the other hand, isolating the `4` makes it a true "helper" index that the section loop never has to handle at all.

src/viperleed/calc/sections/initialization.py
```diff
@@ -601,6 +570,11 @@ def init_domains(rp):
         rp.setHaltingLevel(3)
         return
 
+    if next_segment == 1:
+        # Users explicitly requested a refcalc for all domains.
+        for domain in rp.domainParams:
+            domain.refcalc_required = True
+
```
fkraushofer:
```suggestion
    for idx in rp.RUN:
        # check if refcalcs are explicitly requested, or required but missing
        if idx == 1:
            # Users explicitly requested a refcalc for all domains.
            for domain in rp.domainParams:
                domain.refcalc_required = True
        if idx in (1, 4):
            # a refcalc will be executed if needed, we can stop checking
            break
        if idx in (2, 3, 5) and rr:
            logger.error(
                "Some domains require new reference calculations before "
                "a domain search can be executed. Please either manually "
                "execute appropriate reference calculations, or set RUN = 4"
                )
            rp.setHaltingLevel(3)
            return
```

fkraushofer: The current implementation assumes that if the next section is not a refcalc (or 4), then it is definitely something that requires a refcalc, but this is not true: Things like a full-dynamic optimization (independet of whether it's currently supported for domains), or possibly a superpos or whatever other segments we may add in the future, would also cause the error to be raised. I think more robust would be to check if any of the problematic segments is requested before the next refcalc:
```suggestion
    for idx in rp.RUN:
        # check if refcalcs are explicitly requested, or required but missing
        if idx == 1:
            # Users explicitly requested a refcalc for all domains.
            for domain in rp.domainParams:
                domain.refcalc_required = True
        if idx in (1, 4):
            # a refcalc will be executed if needed, we can stop checking
            break
        if idx in (2, 3, 5) and rr:
            logger.error(
                "Some domains require new reference calculations before "
                "a domain search can be executed. Please either manually "
                "execute appropriate reference calculations, or set RUN = 4"
                )
            rp.setHaltingLevel(3)
            return
```

fkraushofer (changes requested):

michele-riva:
@fkraushofer, would you mind to take a final look?

Issue #23, opened 2021-05-20
Figure out appropriate licensing of ViPErLEED parts


michele-riva:
Before releasing v1.0.0 we have to make sure to have the licenses properly figured out.

The plan is to release the whole ViPErLEED package under "GPL v3 or later". We may also do dual licensing with some license (to be determined) that is compatible with GPL v3, as suggested [here](https://www.gnu.org/licenses/license-compatibility.en.html).

This copyleft licensing level is required as we use PyQt5, [relased under GPL v3](https://riverbankcomputing.com/software/pyqt/intro) (notice that Qt5 is [released under LGPL v3](https://doc.qt.io/qt-5/licensing.html), but GPL v3 [subsumes it](https://www.gnu.org/licenses/license-compatibility.en.html), so we should use GPL v3). See also the [License FAQ](https://riverbankcomputing.com/commercial/license-faq) from Riverbank.

Here are the parts that we are using, including their licenses (where known):
- [x] Python interpreter: [PSF License Agreement](https://docs.python.org/3/license.html#psf-license). [Compatible](https://docs.python.org/3/license.html) with GPL v3 since v2.2.

Python modules:
- [x] fortranformat: [MIT](https://github.com/keweiyao/fortranformat/blob/master/COPYING.txt).  [Subsumed by GPL v3](https://www.gnu.org/licenses/license-compatibility.en.html); also known as [X11](https://en.wikipedia.org/wiki/MIT_License). ~~**Notice**: we are currently distributing a version modified by @fkraushofer, but the `fortranformat` directory does **not** contain the original [LICENSE](https://github.com/keweiyao/fortranformat/blob/master/COPYING.txt). We should make sure we do.~~ Since 3d7ead39ce29569b60c42a2d1dfe24069e5f83d4 we're not vendoring `fortranformat` any longer but depend on it.
- [x] line_profiler: [BSD](https://github.com/rkern/line_profiler/blob/master/LICENSE.txt). [Subsumed by GPL v3](https://www.gnu.org/licenses/license-compatibility.en.html). NB: this module is currenty used only for development.
- [x] matplotlib: [BSD-compatible](https://github.com/matplotlib/matplotlib/blob/master/LICENSE/LICENSE). See also [the matplotlib page discussing GPL vs BSD](https://matplotlib.org/stable/devel/license.html#license-discussion). [Subsumed by GPL v3](https://www.gnu.org/licenses/license-compatibility.en.html).
[...]

PR #468, opened 2025-10-10
Dll from separate repository

FlorianDoerr:
Load drivers from a user-specified location.

For licensing reasons we cannot leave the Imaging Source dlls in the viperleed repository. With the changes from this PR we will be able to load the dlls from a folder.

Takes care of the Imaging Source licensing part of #23
Does not include a download option yet.

Also removes `__init__` imports in controller, camera, and serial, reducing the risk of circular imports.

### DIFF
--- a/tests/gui/measure/classes/test_settings.py
+++ b/tests/gui/measure/classes/test_settings.py
@@ -256,10 +256,9 @@ fallback_values = (('A/opt2', 'cfb'),)
         assert added_section in parser.sections()
 
     @parametrize(cls_name=('IHaveNoAliases', '', None))
-    # pylint: disable-next=unused-argument
     def test_no_aliases(self, cls_name):
         """Check emptiness of aliases when none exist."""
-        parser = AliasConfigParser(cls_name='IHaveNoAliases')
+        parser = AliasConfigParser(cls_name=cls_name)
         assert not parser._aliases      # pylint: disable=protected-access
         assert not parser._fallbacks    # pylint: disable=protected-access
 


src/viperleed/gui/measure/hardwarebase.py
```diff
@@ -45,24 +46,68 @@ def class_from_name(package, class_name):
...
+        raise ValueError(f'No {package} class named {class_name} found.')
+    try:
+        next(classes)
+    except StopIteration:  # Only one module defines class_name: OK
+        return cls
+    raise ImportError(f'More than one sub-module in {package} '
+                      f'defines {class_name} class.')
+
+
+def import_importable_with_sub_modules(importable):
```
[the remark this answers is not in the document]
michele-riva: Perhaps we could already consider the option of making this recursive-enabled by having a `package=None` kwarg that falls back to `__package__` if None (i.e., `package = package or __package__`). If we ever need to import submodules with arbitrary depth, we can later on `all_modules.extend(import_with_submodules(sub_mod.name, package=module.__package__)` within the loop.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    base = 'fakepkg'
+    full_base = 'viperleed.gui.measure.fakepkg'
+
+    pkg = types.ModuleType(base)
+    pkg.__path__ = ['dummy_path']
+    pkg.__package__ = full_base
+
+    sys.modules[base] = pkg
+    sys.modules[full_base] = pkg
```
michele-riva: Rather than writing directly to `sys.modules`, it would be better to use `mocker.patch.dict('sys.modules', {base: pkg, full_base: pkg})` (and the same in `add_submodule`). Otherwise, the editing of `sys.modules` will persist beyond the scope of one test, potentially making tests "talk to one another". Using `mocker.patch` ensures that the added/modified (key,value) pairs are deleted/restored at the end of the test.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    def add_submodule(name, module):
+        sys.modules[f'{base}.{name}'] = module
+        sys.modules[f'{full_base}.{name}'] = module
+        submodules[name] = module
+
+    # Mock importlib.import_module.
+    def fake_import(name, package=None):
+        return sys.modules[name]
+
+    mocker.patch('importlib.import_module', side_effect=fake_import)
```
michele-riva: Do you really need to patch importlib.import_module? I was under the impression that it would fetch from sys.modules if the module is already imported. Since you write to sys.modules, it may be OK not to patch it.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+import types
+import pytest
+
+from viperleed.gui.measure.hardwarebase import class_from_name
+from viperleed.gui.measure.hardwarebase import get_devices
+from viperleed.gui.measure.hardwarebase import import_with_sub_modules
+
+
+@pytest.fixture
+def fake_pkg(mocker):
```
michele-riva:
```suggestion
@pytest.fixture(name='fake_pkg')
def fixture_fake_pkg(mocker):
```

This is to prevent pylint complaints about argument names below shadowing the fake_pkg module-level name.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    A.__module__ = sub_a.__name__
+    sub_a.A = A
+    pkg.add_submodule('sub_a', sub_a)
+
+    return pkg
+
+
+# Tests for import_with_sub_modules
+def test_import_with_sub_modules_yields_main_and_submodules(fake_pkg):
+    result = list(import_with_sub_modules('fakepkg'))
```
michele-riva: You probably do not need to make it into a list since you consume it right below.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    names = [m.__name__ for m in result]
+    assert 'fakepkg' in names
+    assert 'fakepkg.sub_a' in names
```
michele-riva: It would be best to probably test equality rather than inclusion here:

```suggestion
    expect_names = [
        'fakepkg',        # First the top-level module
        'fakepkg.sub_a',  # Then all its submodules
        ]
    assert names == expect_names
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+
+    pkg.add_submodule = add_submodule
+    pkg.submodules = submodules
+
+    # Define submodule containing a class A.
+    sub_a = types.ModuleType('fakepkg.sub_a')
+    class A:
+        pass
+    A.__module__ = sub_a.__name__
+    sub_a.A = A
```
michele-riva:
```suggestion
    sub_a.A = type('A', (), {'__module__': sub_a.__name__})
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    cls = class_from_name('fakepkg', 'A')
+    assert cls.__name__ == 'A'
+
+
+def test_class_from_name_no_class_found(fake_pkg):
+    with pytest.raises(ValueError):
+        class_from_name('fakepkg', 'NotExist')
+
+
+def test_class_from_name_duplicate_classes(fake_pkg):
```
michele-riva: Another test that is missing is the equivalent of what we did manually, i.e., import a class from another submodule into one of the submodules.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, name):
+        self.unique_name = name
+
+
+def test_get_devices_collects_all_devices(fake_pkg):
+    driver_a = types.ModuleType('fakepkg.driver_a')
+
+    class DummyDevice:
+        def __init__(self, name):
+            self.unique_name = name
```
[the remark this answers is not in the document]
michele-riva: Also, perhaps you want to call this a `DummyDeviceInfo`? Then `DummyDeviceClass` can become `DummyDevice`.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    class DummyDeviceClass:
+        def list_devices(self):
+            return [DummyDevice('device_a')]
+
+    DummyDeviceClass.__module__ = driver_a.__name__
+    driver_a.DummyDeviceClass = DummyDeviceClass
+    fake_pkg.add_submodule('driver_a', driver_a)
+
+    devices = get_devices('fakepkg')
+    assert 'device_a' in devices
```
michele-riva: Here you know that there must be only a single device, so I'd also check len(devices) == 1

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+    class DummyDevice:
+        def __init__(self, name):
+            self.unique_name = name
+
+    class DummyDeviceClass:
+        def list_devices(self):
+            return [DummyDevice('device_a')]
+
+    DummyDeviceClass.__module__ = driver_a.__name__
+    driver_a.DummyDeviceClass = DummyDeviceClass
```
michele-riva: Since you do this multiple times, you could make a decorator for the class:

```python
def in_module(module):
    """Register a class to belong to a module."""
    def _register_cls(cls):
        cls.__module__ = module.__name__
        setattr(module, cls.__name__, cls)
        return cls
    return _register_cls
```

To be used as

```python
@in_module(driver_a)
class DummyDeviceClass:
    def list_devices(self):
        ...
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,156 @@
...
+        @abc.abstractmethod
+        def list_devices(self):
+            pass
+
+    AbstractDevice.__module__ = sub.__name__
+    sub.AbstractDevice = AbstractDevice
+    fake_pkg.add_submodule('drivers', sub)
+
+    devices = get_devices('fakepkg')
+    assert devices == {}
```
michele-riva:
```suggestion
    assert not devices
```

michele-riva (dismissed):
Some suggestions for the tests

tests/gui/measure/_defaults/test_defaults.py
```diff
@@ -37,6 +37,7 @@ def list_default_settings():
 class TestDefaultAliases:
     """Tests for the aliases provided in the _defaults."""
 
+    # pylint: disable=magic-value-comparison
```
michele-riva: If you move this inside the function body, you don't need to re-enable it afterwards

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+    submodules = {}
+
+    def add_submodule(name, module):
+        """Add submodule to fake package."""
+        mocker.patch.dict('sys.modules', {f'{base}.{name}': module,
+                                          f'{full_base}.{name}': module})
+        submodules[name] = module
+
+    # Mock pkgutil.iter_modules to reflect whatever submodules exist.
+    def fake_iter_modules(path):
```
michele-riva: Here, since `path` is unused (but you need it to match the signature of iter_modules), you can replace it with `_`. Then pylint will not complain.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+class TestImportWithSubModules:
+    """Tests for the import_with_submodules iterator."""
+
+    def test_invalid_import(self, mocker):
+        """Test import error."""
+        mocker.patch('importlib.import_module', side_effect=ImportError)
+        with pytest.raises(AttributeError):
+            _ = list(import_with_sub_modules('nonexistent'))
+
+    def test_yield_main_and_submodule(self, fake_pkg):
```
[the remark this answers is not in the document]
michele-riva: Do the same for the other tests that need it for setup/teardown, but don't use the argument explicitly. Since you will use this multiple times, it may be worth defining a module-level "shortcut":

```python
with_fake_pkg = pytest.mark.usefixtures('fake_pkg')

...

    @with_fake_pkg
    def test_that_needs_fake_pkg_but_not_as_an_argument(self):
        ...
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+        """Test finding class failure."""
+        with pytest.raises(ValueError):
+            class_from_name('fakepkg', 'NotExist')
+
+    def test_duplicate_classes(self, fake_pkg):
+        """Test duplicate class declaration."""
+        # Add submodule that defines another DummyA
+        sub_b = types.ModuleType('fakepkg.sub_b')
+        @in_module(sub_b)
+        class DummyA: # pylint: disable=too-few-public-methods, unused-variable
```
michele-riva:
```suggestion
        class DummyA:  # pylint: disable=too-few-public-methods, unused-variable
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+
+        cls = class_from_name('fakepkg', 'DummyA')
+        assert cls is sub_a.DummyA
+        assert cls.__module__ == sub_a.__name__
+
+
+class TestGetDevices:
+    """Tests for the get_devices function."""
+
+    def test_device_getting(self, fake_pkg):
```
michele-riva:
```suggestion
    def test_get_device(self, fake_pkg):
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def test_ignore_non_device_classes(self, fake_pkg):
+        """Test ignoring non-devices."""
+        sub = types.ModuleType('fakepkg.drivers')
+
+        # Add abstract class with list_devices. 'DummyA' already is
+        # a non-abstract class without 'list_devices', so we do not
+        # need to add a non-device class.
+        @in_module(sub)
+        # pylint: disable=too-few-public-methods, unused-variable
```
michele-riva:
```suggestion
        # pylint: disable-next=too-few-public-methods, unused-variable
```

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+import types
+
+import pytest
+
+from viperleed.gui.measure.hardwarebase import class_from_name
+from viperleed.gui.measure.hardwarebase import get_devices
+from viperleed.gui.measure.hardwarebase import import_with_sub_modules
+
+
+# pylint: disable=magic-value-comparison, unused-argument
```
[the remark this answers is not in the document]
michele-riva: To prevent magic-value-comparisons, it may actually be better to define (module globals or) local variables with string values to compare against. Then, if you ever need to change the value, there is only one spot to modify.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,210 @@
...
+@pytest.fixture(name='fake_pkg')
+def fixture_fake_pkg(mocker):
+    """Create a fake package structure like:
+    fakepkg/
+        sub_a.py
+    and mock sys.modules and pkgutil.iter_modules so that
+    import_with_sub_modules and related functions can discover them.
+    """
+    base = 'fakepkg'
+    full_base = 'viperleed.gui.measure.fakepkg'
```
michele-riva:
```suggestion
    full_base = f'viperleed.gui.measure.{base}'
```

tests/gui/measure/test_hardwarebase.py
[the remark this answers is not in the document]
michele-riva: (NB: the tmp_path fixture that you were not using in other tests before does not need to be included in the usefixtures mark. If you're not using it, the test will create a temporary directory, but will still execute in the the path in which you called pytest.)

michele-riva (changes requested):
Some small suggestions for the tests

tests/gui/measure/_defaults/test_defaults.py
```diff
@@ -115,7 +116,7 @@ def test_no_multiple_inheritance(self, alias_config):
                     )
 
 
-class TestDefaultSettings:
+class TestDefaultSettings: # pylint: disable=too-few-public-methods
```
michele-riva:
```suggestion
class TestDefaultSettings:  # pylint: disable=too-few-public-methods
```

tests/gui/measure/classes/test_settings.py
```diff
@@ -258,11 +256,12 @@ def test_new_sections_added(self):
         assert added_section in parser.sections()
 
     @parametrize(cls_name=('IHaveNoAliases', '', None))
+    # pylint: disable-next=unused-argument
     def test_no_aliases(self, cls_name):
         """Check emptiness of aliases when none exist."""
         parser = AliasConfigParser(cls_name='IHaveNoAliases')
```
michele-riva: Here the unused-argument actually signals a problem: The `cls_name` argument should be used instead of `'IHaveNoAliases'` for initializing `AliasConfigParser`. Otherwise the parametrization is useless.

tests/gui/measure/test_hardwarebase.py
```diff
@@ -0,0 +1,219 @@
...
+        # Import DummyA from sub_a into sub_b
+        sub_a = fake_pkg.submodules['sub_a']
+        sub_b = types.ModuleType('fakepkg.sub_b')
+
+        # Re-export DummyA (imported reference)
+        sub_b.DummyA = sub_a.DummyA
+        fake_pkg.add_submodule('sub_b', sub_b)
+
+        cls = class_from_name('fakepkg', _DUMMY_NAME)
+        assert cls is sub_a.DummyA
```
michele-riva:
```suggestion
        assert cls is imported_cls
```

michele-riva (approved):
Good to go!

PR #470, opened 2025-10-16
Check system settings valid

### LINKED ISSUE #403: Measure UI: no complaints if system settings paths do not exist
Steps to reproduce:
- create some folders for the system settings
- select them, save settings
- close software
- delete folders
- restart software

Expected behavior: the software complains that the paths do not exist.

This also means that Devices>Camera>pick one crashes.

FlorianDoerr:
System settings are checked for validity, given paths now must be directories. System settings validity check is enforced before devices are opened and cannot be dismissed.

Fixes #403

### DIFF
--- a/src/viperleed/gui/measure/widgets/pathselector.py
+++ b/src/viperleed/gui/measure/widgets/pathselector.py
@@ -146,7 +146,7 @@ class PathSelector(qtw.QWidget):
         self.__lineedit.setReadOnly(self.read_only)
 
         # Add/remove the spacer item between line edit and browse
-        # It is always at position 1: '0' is
+        # It is always at position 1, 0 is self.__lineedit.
         layout = self.layout()
         if self.read_only:
             layout.removeItem(layout.itemAt(1))


src/viperleed/gui/measure/classes/settings.py
```diff
@@ -847,6 +842,17 @@ def get_settings_handler(self):
...
 
+    def valid(self):
+        """Return whether settings are valid and a reason if invalid."""
+        if not all(self[sec][opt] for sec, opt in self.__non_null):
+            return (False, 'Missing mandatory (*) settings. Please fill '
+                           'in all mandatory fields in the next dialog.')
+        for folder in self['PATHS'].values():
+            if not Path(folder).is_dir():
+                return (False, 'Invalid path detected. Please make sure that '
+                               'the given paths only point to directories.')
```
michele-riva: Perhaps it may be good to edit the `PathSelector` widget to clearly show offending paths. E.g., it may have a .check_path_exists() method that turns the text red/black depending on whether the path is a valid one. (It can then be more specific as to whether the path should be a file or a directory depending on what it allows to select.) You would then have to iterate over all of them here, to ensure all the problematic ones are marked as such.

src/viperleed/gui/measure/widgets/pathselector.py
```diff
@@ -144,7 +146,7 @@ def read_only(self, read_only):
         self.__lineedit.setReadOnly(self.read_only)
 
         # Add/remove the spacer item between line edit and browse
-        # It is always at position 1: "0" is
+        # It is always at position 1: '0' is
```
michele-riva: You're right, this comment looks like it's missing a part. The item at position 0 is self.__lineedit.

src/viperleed/gui/measure/widgets/pathselector.py
```diff
@@ -155,6 +157,14 @@ def set_path(self, new_path):
         """Set the contents of this path selector."""
         self.__glob['full_path'] = Path(new_path)
         self.__lineedit.setToolTip(str(new_path))
+        if self.__glob['select_file']:
+            valid = self.__glob['full_path'].is_file()
+            if not self.__glob['existing_file']:
+                valid = not valid
+        else:
+            valid = self.__glob['full_path'].is_dir()
```
michele-riva:
```suggestion
        full_path = self.__glob['full_path']
        selects_file = self.__glob['select_file']
        if not selects_file:
            valid = full_path.is_dir()
        elif selects_file and self.__glob['existing_file']:
            valid = full_path.is_file()
        else:  # A file to be selected for saving
            valid = True
```

A bit nicer to have a flat if...elif...else. Also, I think when a PathSelector is meant to select a file for saving, it does not matter whether the file exists or not. The assumption is that it will be created or overwritten.

michele-riva (changes requested):
Some small suggestions

Issue #404, opened 2025-06-06
Uninitialized controller crashes UI


michele-riva:
Steps to reproduce:
- open UI
- pick a settings folder missing a file for a controller
- New Measurement, select that controller

Result:
```
Traceback (most recent call last):
  File "...\src\viperleed\gui\measure\widgets\collapsiblelists.py", line 214, in _get_relative_path
    path = path.relative_to(self.default_settings_folder)
AttributeError: 'NoneType' object has no attribute 'relative_to'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "...\src\viperleed\gui\measure\measurement\_meassettings.py", line 146, in _store_device_settings
    str(self._controllers.get_primary_settings()))
  File "...\src\viperleed\gui\measure\widgets\collapsiblelists.py", line 438, in get_primary_settings
    rel_path = self._get_relative_path(view.settings_file)
  File "...\src\viperleed\gui\measure\widgets\collapsiblelists.py", line 216, in _get_relative_path
    new_path = str(path.as_posix())
AttributeError: 'NoneType' object has no attribute 'as_posix'
```

PR #471, opened 2025-10-16
fix: refusing to set camera settings

FlorianDoerr:
When refusing to set camera settings, the measurement UI used to break.


FlorianDoerr:
> I guess this is the equivalent of #404 for cameras

Yes

michele-riva (approved):

PR #475, opened 2025-10-20
Fix device settings dialog without settings

### LINKED ISSUE #392: List of controller settings not up to date with config folder
**Steps to reproduce:**
- Open GUI
- Open a controller settings dialog
- Close it
- Delete its settings file from the configuration folder
- Open again the same controller, without closing the UI

--> The former settings dialog opens up

Reason: we keep alive settings (or even settings dialogs) of known devices.
Solution: check that the settings file the dialog refers to still exists. If not, ask users what to do: restore the settings or start fresh?

FlorianDoerr:
Catch if the device settings file has been deleted by the time a device settings dialog is created.

Fixes #392

### DIFF
--- a/src/viperleed/gui/measure/uimeasurement.py
+++ b/src/viperleed/gui/measure/uimeasurement.py
@@ -1073,7 +1073,7 @@ class Measure(ViPErLEEDPluginBase):                                             
         self.system_settings.read_again()
         # Note that we are using exec() and an immediate call to
         # _check_sys_settings_ok() to create a loop and force the user
-        # to give valid settings.
+        # to give valid settings. QDialog.open would not block.
         self._dialogs['sys_settings'].exec()
         self._check_sys_settings_ok()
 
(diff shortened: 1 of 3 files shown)


src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -876,6 +880,13 @@ def _on_controller_clicked(self, *_):
         if not _dialog:
             # Something went wrong with creating the dialog
             return
+        if not Path(_dialog.settings.last_file).is_file():
```
michele-riva:
```suggestion
        if not _dialog.settings.last_file.is_file():
```

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -876,6 +880,13 @@ def _on_controller_clicked(self, *_):
...
+            # Old settings dialog and file is missing.
```
michele-riva:
```suggestion
            # Old settings dialog and file is missing (see #392).
```

michele-riva (changes requested):
Some small suggestions

michele-riva (approved):

PR #477, opened 2025-10-22
Move handling of unsent messages to serial

### LINKED ISSUE #157: Move queuing of commands to serial
Commands that cannot be sent because the serial is busy are stored in a queue on the controller side. Unsent messages should be stored on the serial side instead. When implementing this, the __force_stop_timer has to be started on the controller side after a stop command has been sent to the serial. If this timer times out, then the unsent messages have to be cleared and the private __busy property of the serial has to be set to false (private as to avoid emitting the serial_busy signal). Then the stop command must be sent again.

### LINKED ISSUE #334: Move _unsent_messages handling to serial level
`_unsent_messages` are currently handled by the controller. This was most likely necessary due to a bug with the threading implementation. We fixed this bug since then and should therefore move handling of `_unsent_messages` to the `SerialABC` class.

FlorianDoerr:
Move storing and sending of unsent messages to serial.

Fixes #157
Fixes #334

### DIFF
--- a/src/viperleed/gui/measure/controller/abc.py
+++ b/src/viperleed/gui/measure/controller/abc.py
@@ -924,12 +924,6 @@ class ControllerABC(DeviceABC):
             sent. (Regardless of whether the stop command has been sent
             before or during the execution of .force_stop().)
         """
-        self.serial.unsent_messages.clear()
-        with disconnected_signal(self.serial.busy_changed,
-                                 self.__do_preparation_step,
-                                 self.set_busy,
-                                 type=_UNIQUE):
-            self.serial.busy = False
         if not self.settings or not self.connected:
             # Settings missing or serial no longer connected.
             return False
@@ -937,6 +931,12 @@ class ControllerABC(DeviceABC):
             # The timer was started, but the serial managed
             # to send the stop command in the meanwhile.
             return False
+        self.serial.unsent_messages.clear()
+        with disconnected_signal(self.serial.busy_changed,
+                                 self.__do_preparation_step,
+                                 self.set_busy,
+                                 type=_UNIQUE):
+            self.serial.busy = False
         return True
 
     def true_energy_to_setpoint(self, energy):


src/viperleed/gui/measure/controller/abc.py
```diff
@@ -241,9 +223,7 @@ def set_busy(self, is_busy):
         busy_changed
             If the busy state of the controller changed.
         """
-        if self._unsent_messages:
-            return
-        if self.serial and self.serial.busy:
+        if self.serial and (self.serial.busy or self.serial.unsent_messages):
```
michele-riva: Perhaps we should consider the serial busy if it has unsent messages?

src/viperleed/gui/measure/controller/abc.py
```diff
@@ -683,7 +653,7 @@ def disconnect_(self):
 
     def flush(self):
         """Clear unsent messages and set not busy."""
-        self._unsent_messages = []
+        self.serial.unsent_messages = []
```
michele-riva:
```suggestion
        self.serial.unsent_messages.clear()
```

src/viperleed/gui/measure/controller/abc.py
```diff
@@ -948,19 +891,24 @@ def set_measurements(self, quantities):
...
-        base.safe_connect(serial_busy, self.send_unsent_messages,
-                          type=_QUEUED_UNIQUE)
         base.safe_connect(serial_busy, self.set_busy, type=_QUEUED_UNIQUE)
-        self.__force_stop_timer.start()
+
+    @abstractmethod
+    @qtc.pyqtSlot()
+    def force_stop(self):
+        """Force the controller to stop."""
+        self.serial._busy = False
```
michele-riva: Not great to access a private member of the serial. Perhaps `SerialABC` should itself have a `force_stop` method that is called in here?

src/viperleed/gui/measure/controller/viperinocontroller.py
```diff
@@ -1065,6 +1065,24 @@ def stop(self):
...
+    @qtc.pyqtSlot()
+    def force_stop(self):
+        """Force the controller to stop."""
+        if not self.settings or not self.connected:
+            # Settings missing or serial no longer connected.
+            return
+        if not self.serial.unsent_messages:
+            # The timer was started, but the serial managed
+            # to send the stop command in the meanwhile.
+            return
```
michele-riva: It feels like these could be more general than just the viperino. Perhaps they should be in the super()? In principle, they could both go to the SerialABC implementation.

src/viperleed/gui/measure/serial/abc.py
```diff
@@ -161,7 +161,23 @@ def __init__(self, settings, port_name='', **kwargs):
...
+        # in the order they arrive at. When the serial receives
+        # a valid answer from the hardware it will automatically
+        # trigger an attempt to send the next message in
+        # unsent_messages. Each element of unsent_messages is
+        # a tuple whose first element is the command and its associated
+        # data, and the second element is the timeout parameter.
+        self.unsent_messages = []
+
         self.__move_to_thread_requested.connect(self.__on_moved_to_thread)
+        _qu = qtc.Qt.QueuedConnection | qtc.Qt.UniqueConnection
```
michele-riva: This can be a module-level constant

michele-riva (changes requested):
Some suggestions, like always.

src/viperleed/gui/measure/controller/abc.py
michele-riva: Let's leave it as is, seems that this would make it too complex

src/viperleed/gui/measure/controller/viperinocontroller.py
```diff
@@ -1065,6 +1065,26 @@ def stop(self):
...
+    @qtc.pyqtSlot()
+    def force_stop(self):
+        """Force the controller to stop.
+
+        Returns
+        -------
+        must_stop : bool
+            In this reimplementation this will always be False. Only
+            required to match the behaviour of super().force_stop().
+        """
```
michele-riva:
```suggestion
        """Force the controller to stop."""
```

src/viperleed/gui/measure/controller/viperinocontroller.py
```diff
@@ -1065,6 +1065,26 @@ def stop(self):
...
+        if not super().force_stop():
+            return False
+        stop = self.settings.get('available_commands', 'PC_STOP')
+        self.send_message(stop)
+        return False
```
michele-riva:
```suggestion
        return False  # Already re-sent STOP
```

michele-riva (changes requested):
Some more suggestions

src/viperleed/gui/measure/controller/abc.py
```diff
@@ -943,24 +887,55 @@ def set_measurements(self, quantities):
...
+        still has to be sent.
+
+        Returns
+        -------
+        must_stop : bool
+            Whether the controller still has to be stoppped. If True,
+            another stop command has to be sent. Subclasses should only
+            return False unless they are subclassed themselves.
+        """
+        self.serial.unsent_messages.clear()
```
michele-riva: Perhaps you should also check `if not self.serial` before this. Otherwise we risk an `AttributeError`.

michele-riva (changes requested):
Some final small suggestions

src/viperleed/gui/measure/controller/abc.py
FlorianDoerr: This check is already performed in
```
        if not self.settings or not self.connected:
            # Settings missing or serial no longer connected.
            return False
 ```
 as connected returns:
 `return self.serial and self.serial.is_open`

PR #480, opened 2025-10-28
Fix disconnect during bad pixel calibration

### LINKED ISSUE #381: Disconnecting camera during bad pixel calibration
Disconnecting camera during bad pixel calibration raises exception. We should report a non-breaking error here.

FlorianDoerr:
Fixes #381


src/viperleed/gui/measure/camera/drivers/imagingsource/tisgrabber.py
```diff
@@ -971,6 +982,88 @@ def pause(self):
...
+        for cb in (on_frame_ready, on_disconnected):
+            if not hasattr(cb, '__call__'):
+                raise TypeError(f'Callback {cb.__name__} '
+                                'is not a callable.')
+            if (not hasattr(cb, '__ctypeswrapper__')
+                    or cb.__ctypeswrapper__ not in allowed):
+                raise ValueError(f'Callback {cb.__name__} was not '
+                                 'decorated with an allowed callback '
+                                 'type. This is needed to ensure '
+                                 'appropriate type checking/conversions')
```
michele-riva: You could factor out the code in `set_frame_ready_callback` to make it accept a `func` and a `wrapper_marker` arguments. Then you can reuse it in both methods. Here I'd suggest to use `zip` to catch the correct `__ctypeswrapper__` for each callback rather than using `in allowed`.

src/viperleed/gui/measure/camera/drivers/imagingsource/tisgrabber.py
```diff
@@ -1017,6 +1110,8 @@ def set_frame_ready_callback(self, on_frame_ready, py_obj_for_callback):
             If this method is called more than once
             before rebooting a camera.
         """
+        raise RuntimeError('This function should no longer be used.'
```
michele-riva: `NotImplementedError`

michele-riva (changes requested):
Some suggestions to the current implementation as you work out the other issues.

michele-riva (changes requested):
Some more suggestions while you're still working out the rest

FlorianDoerr:
Converting this to a draft as it will still need a lot more work.

PR #481, opened 2025-10-29
fix: create global settings folder

FlorianDoerr:
Create global settings folder if it does not exist yet.


FlorianDoerr:
> Looks good. Let's remember to add a test for this when we have tests for the ViPErLEEDSettings class.

Currently looking into this.

PR #483, opened 2025-10-30
fix,tests: more failing symmetry tests

michele-riva:


michele-riva:
> Fine to add of course, but do we understand why these fail? I never looked at these tests before I think.

There are a bunch of (small) problems in the symmetry finding code for which I have fixes in #51. Unfortunately I still have to finish that one.

PR #485, opened 2025-11-05
Implement requests download fallback

FlorianDoerr:
Use requests module as fallback in case QNetwork cannot be used due to issues with OpenSSL.

(PyQt5 was built against OpenSSL 1.1 and some newer operating systems like Ubuntu 24.04 no longer support this OpenSSL version.)


src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,190 @@
+"""Module downloader of viperleed.gui.measure.classes.
+
+Contains NetworkHelper, a downloader that uses QNetwork or requests as
```
michele-riva:
```suggestion
Contains NetworkHelper, a downloader that uses QtNetwork or requests as
```

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,190 @@
...
+__authors__ = (
+    'Michele Riva (@michele-riva)',
+    'Florian Dörr (@FlorianDoerr)',
+    )
+__copyright__ = 'Copyright (c) 2019-2025 ViPErLEED developers'
+__created__ = '2025-11-04'
+__license__ = 'GPLv3+'
+
+import importlib
+import typing
```
michele-riva: Remove `typing` and all type hints below

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,190 @@
...
+    """Abstraction for QNetwork with requests fallback."""
+
+    # Emitted if the download was successful. Carries downloaded data.
+    download_finished = qtc.pyqtSignal(bytes)
+
+    # Emitted if the download failed.
+    download_failed = qtc.pyqtSignal()
+
+    # Emitted if the requests module must be installed. This is usually
+    # the case if Qt5 QNetwork is incompatible with the OS.
```
michele-riva:
```suggestion
    # the case if Qt5 QtNetwork is incompatible with the OS.
```

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,190 @@
...
+    install_requests = qtc.pyqtSignal()
```
michele-riva: Maybe `requests_module_not_found`? Or, even better, if we should ever extend this, `fallback_not_found(str)`.

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,190 @@
...
+        None.
+        """
+        if not qtn.QSslSocket.supportsSsl():
+            # Cannot use QNetwork, use requests module instead.
+            self._download_with_requests(url)
+            return
+
+        request = qtn.QNetworkRequest(qtc.QUrl(url))
+        # Since Qt does not come with SSL support the RedirectAttribute
+        # must be set in order to get the Arduino CLI file via http.
```
michele-riva: Could you verify this comment? Looks like Qt does have SSL support

src/viperleed/gui/measure/dialogs/firmwareupgradedialog.py
```diff
@@ -541,6 +542,25 @@ def _on_controllers_detected(self, data_dict):
...
+            'It seems your system cannot use the built-in Qt network '
+            'support. To continue, you need to install the Python '
+            '<b>requests</b> module in the environment in which you '
+            'have installed ViPErLEED. You can do this in your terminal'
+            ' or command prompt by activating your environment and '
+            'excuting the following command:'
+            '<p><center><b>python -m pip install requests</b></center></p>'
+            'Restart ViPErLEED after installing <b>requests</b>.'
+            )
+        notify.addButton(qtw.QPushButton('Ok'), notify.AcceptRole)
```
michele-riva:
```suggestion
        notify.addButton(notify.Ok)
```

michele-riva (changes requested):
Some suggestions like always

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,190 @@
...
+            requests = importlib.import_module('requests')
+        except ImportError:
+            self.failed.emit()
+            return
+        try:
+            response = requests.get(self._url, timeout=2)
+            response.raise_for_status()
+            self.finished.emit(response.content)
+        except requests.exceptions.RequestException:
+            self.failed.emit()
```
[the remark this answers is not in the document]
FlorianDoerr: Both of them can raise, I originally had the same thought.

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,189 @@
...
+            return
+        self._thread = qtc.QThread(self)
+        self._worker = _RequestsWorker(url)
+        self._worker.moveToThread(self._thread)
+
+        # Wire signals
+        self._worker.finished.connect(self._cleanup_thread)
+        self._worker.failed.connect(self._cleanup_thread)
+        self._worker.finished.connect(self.download_finished.emit)
+        self._worker.failed.connect(self.download_failed.emit)
```
michele-riva:
```suggestion
        self._worker.finished.connect(self.download_finished)
        self._worker.failed.connect(self.download_failed)
```

I think that signals can be connected to signals directly.

michele-riva (changes requested):
Some very tiny suggestions left, but looks very good.

src/viperleed/gui/measure/classes/downloader.py
```diff
@@ -0,0 +1,189 @@
...
+    """Downloder using requests module."""
+
+    # Emitted if download finished successfully.
+    finished = qtc.pyqtSignal(bytes)
+    # Emitted if download failed.
+    failed = qtc.pyqtSignal()
+
+    def __init__(self, url):
+        """Initialise downloader."""
+        super().__init__()
```
michele-riva:
```suggestion
    def __init__(self, url, parent=None):
        """Initialise downloader."""
        super().__init__(parent=parent)
```

Just to maintain the same interface as the `QObject` base class.

michele-riva (changes requested):
One last tiny suggestion

michele-riva (approved):
Looks good!

PR #489, opened 2025-11-08
v0.14.2

michele-riva:


michele-riva:
Will bump the minor instead (to [v0.15.0](https://github.com/viperleed/viperleed/pull/541))

PR #492, opened 2025-11-11
Robust R factor selection

alexmiame:
This PR implements a new, more robust R factor selection and adds the option to use the new smooth R factor with the vipeleed-jax plugin.
As I wanted to add the option for the new R factor, I noticed that the current selection is rather basic and only supported integer selection of R factors.


alexmiame:
This probably should also include a small update to the `R_FACTOR_TYPE` doc page.

@fkraushofer Feel free to try this and let me know.

fkraushofer:
I added some basic documentation in 35f8d25babbbbb7cb7630d2699e0e5269a5416ae - would be good if @amimre and @schmid-iap could have a quick look to make sure I didn't misrepresent anything. This should be fleshed out more once we add support for the TensErLEED-based R-factors.

PR #497, opened 2025-11-17
Fix bug in writing fd optimization beams

fkraushofer:
If beams have been removed from IVBEAMS, the refcalc automatically drops corresponding beams from EXPBEAMS. In the full-dynamic optimization, this change is currently not propagated back to the "main" rp object. This causes write_fd_opt_beams_pdf to fail because the expbeams labels don't match up with the actually calculated data any more.


alexmiame (approved):
Looks good

PR #499, opened 2025-11-18
No error for requested deltas in JAX backend

fkraushofer:
Ignore request for deltas if e.g. RUN=1-3, only warn, as discussed in #331. However, *do* raise an error if error calculations are requested, since those are actually not implemented and would fail with a non-helpful error message.


fkraushofer:
Just tested this in a "real" run, works as intended.

alexmiame (approved):
Nice!

PR #503, opened 2025-11-28
Disable visual ROI selection

FlorianDoerr:
Visual ROI selection is bugged, so it will be disabled for now. This will be undone once we fixed the ROI.
Fix bug when using None as ROI value. (Would not apply full ROI the first time the CameraViewer was opened with None as ROI value.)

### DIFF
--- a/src/viperleed/gui/measure/camera/abc.py
+++ b/src/viperleed/gui/measure/camera/abc.py
@@ -512,9 +512,9 @@ class CameraABC(DeviceABC):
         Returns
         -------
         tuple
-            Returns an empty tuple if the value in the settings is
-            'None' or if any error occurred while trying to interpret
-            the setting.
+            Returns an empty tuple if any error occurred while trying to
+            interpret the setting, or the full ROI if the value in the
+            settings is 'None'.
 
             Otherwise the following elements are returned:
             roi_x, roi_y : int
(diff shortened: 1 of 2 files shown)


michele-riva (changes requested):
If I read the code correctly, __on_roi_settings_changed does make the ROI rectangle visible once users modify the values in the settings dialog. This feels correct, as we want users to see the ROI visually, but (i) you may want to check that this does not bring back the problems (even if people do not touch the rectangle), and, if (i) is OK, we should probably disable graphical editing of the rectangle altogether.

FlorianDoerr:
> If I read the code correctly, __on_roi_settings_changed does make the ROI rectangle visible once users modify the values in the settings dialog. This feels correct, as we want users to see the ROI visually, but (i) you may want to check that this does not bring back the problems (even if people do not touch the rectangle), and, if (i) is OK, we should probably disable graphical editing of the rectangle altogether.

You are right, it brings the problem back. I disabled it now for this PR.

michele-riva (changes requested):
I fear that some of the changes in cameraviewer.py will go "unnoticed". Perhaps you could have a module global `ROI_RECT_ACTIVE = False` to be used in conditional blocks rather than commenting out sections of code? Then it would become easy to spot the related changes when we fix the real problem.

michele-riva (approved):
Looks good!

PR #504, opened 2025-12-23
Debug new DISPLACEMENTS parser

fkraushofer:
Fixes some issues with the new DISPLACEMENTS parser:
- [x] The regex replacing whitespace-separated matrices for linear operation parsing would not catch e.g. "[1 -1]" because the lookahead did not account for the possibility that the next number may begin with a sign. Fixed and added tests for such cases.
- [x] Linking two atoms with the linear operation "[[-1 0 0] [0 1 0] [0 0 1]]" currently interprets the first coordinate as z, since coordinates are internally ordered as (zxy). However, in input files, we generally follow (xyz), so this should be internally reshuffled.


fkraushofer:
The second point turned out to be better addressed in the viperleed-jax repository, see: https://github.com/viperleed/viperleed-jax/pull/17

fkraushofer:
Seems to be working, ready to merge imo

alexmiame (approved):

PR #509, opened 2026-02-18
Remove negative pixel badness

FlorianDoerr:
Remove the -1 in the terms for the badness calculation as negative badnesses in combination with pixel weights would lead to worse pixels being preferred if they are further away.

### DIFF
--- a/src/viperleed/gui/measure/camera/badpixels.py
+++ b/src/viperleed/gui/measure/camera/badpixels.py
@@ -6,6 +6,7 @@ from a camera and computes a pixel-badness array.
 
 __authors__ = (
     'Michele Riva (@michele-riva)',
+    'Florian Dörr (@FlorianDoerr)',
     )
 __copyright__ = 'Copyright (c) 2019-2025 ViPErLEED developers'
 __created__ = '2021-10-09'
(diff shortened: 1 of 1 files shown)


FlorianDoerr:
image

Here a comparison between the old and the new bad pixel calibration.

michele-riva:
Quick comment before any more progress is done here. Perhaps it would be easier to keep the -1 bits and rather modify the badness at the end when it is scaled by the weights? There one would do `(badness+MIN_BADNESS)*weights` where `MIN_BADNESS` is a module-level constant that depends on the various terms that enter the badness summation (1 for each of the terms, except the long-term telegraph noise that counts for 1.5 --- though this may change if we modify the term from `(x-1)**2` to `x**2-1`). Then all the intermediate thresholds would remain unchanged to a value of 6, which, I think, makes it easier to follow the numbers rather than having to adjust the threshold depending on how many terms have been added so far.

FlorianDoerr:
> Quick comment before any more progress is done here. Perhaps it would be easier to keep the -1 bits and rather modify the badness at the end when it is scaled by the weights? There one would do `(badness+MIN_BADNESS)*weights` where `MIN_BADNESS` is a module-level constant that depends on the various terms that enter the badness summation (1 for each of the terms, except the long-term telegraph noise that counts for 1.5 --- though this may change if we modify the term from `(x-1)**2` to `x**2-1`). Then all the intermediate thresholds would remain unchanged to a value of 6, which, I think, makes it easier to follow the numbers rather than having to adjust the threshold depending on how many terms have been added so far.

Michael told me to get rid of the -1s.

FlorianDoerr:
Deleting this branch as we noticed that scaling multiplicatively with the pixel weights can lead to very bad pixels being favoured over good pixels.

PR #510, opened 2026-02-19
Fix pixel badness scaling

FlorianDoerr:
- Fix the bad pixel scaling that previously preferred worse pixels that were at greater distance from the pixel to replace. This was the case for better-than-average pixels (i.e., those with badness < 0). For example: a worse next-nearest-neighbor  (weight=2) with badness = -0.2 would be picked over a better nearest neighbor (weight=1) with badness = -0.35 because of the **multiplicative** nature of the weighing. Now distance-dependent badness is calculated from the raw one by **addition** of the same weights.
- In the telegraph noise detection, replace `((flicker/flicker_mean - 1)**2) * 1.5` with `((flicker/flicker_mean)**2 - 1) * 0.75`. The former would modify pixels that are better than the average (i.e., those with `flicker/flicker_mean < 1`) and give them a worse score (`>0`) than those at the average (`==0`). This is because the contribution to the badness was not monotonic in the relevant range (`flicker/flicker_mean > 0`). The new term is monotonic, instead: it is negative for better-than-average pixels and positive for worse-than-average ones, as expected.

### DIFF
--- a/src/viperleed/gui/measure/camera/badpixels.py
+++ b/src/viperleed/gui/measure/camera/badpixels.py
@@ -560,7 +560,7 @@ class BadPixelsFinder(_calib.CameraCalibrationTask):
         picked up by each pixel. The threshold when a pixel is
         considered to have too much telegraph noise is when the
         flicker of the pixel is about 3 times the average flicker.
-        (3^2 - 1)* 0.75 = 6,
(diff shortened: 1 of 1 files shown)


michele-riva (approved):
Looks good!

PR #511, opened 2026-02-23
Fix measurement backwards compatibility

FlorianDoerr:
1) New aliases overwrite old aliases.
2) Create suffix and comment keys in outdated settings files.


michele-riva (approved):
Looks good

PR #519, opened 2026-03-17
Implement energy ramp

### LINKED ISSUE #513: Energy generator class
Implement an energy generator class that provides the next energy for a measurement. This class will allow us to handle negative energy steps (down ramps), endless measurements that perform up and down steps indefinitely, and other irregular step profiles without needing to code them into the measurement classes.

FlorianDoerr:
The progression of the energy is now handled by EnergyRamp classes.

Energy ramps are automatically determined from the settings. Energy progression is no longer dependent on the measurement class. Energy ramps support negative delta energies.

Closes #513

### DIFF
--- a/src/viperleed/gui/measure/_defaults/_time_resolved.ini
+++ b/src/viperleed/gui/measure/_defaults/_time_resolved.ini
@@ -24,6 +24,8 @@ measurement_interval = 400
 [energies]
 # Settings of the energy ramp of the measurement
 # ----------------------------------------------
+# energy_ramp
+#   The energy ramp type the measurement will follow.
 # start_energy
 #   The energy at which the time-resolved measurement will start.
 # delta_energy
@@ -39,21 +41,11 @@ measurement_interval = 400
 #   means current energy). Also, fraction = 1 should not be explicitly
 #   included as the last step (nor should its settle time), as this is
 #   added automatically.
-# endless
-#   Whether the measurement should be endless or not. If True, the
-#   time-resolved measurement will return to start_energy upon reaching
-#   end_energy. To end an endless measurement one has to abort the
-#   measurement.
-# constant_energy
-#   Whether the energy should be incremented at each step or not. If
-#   True the energy will never deviate from the start_energy, which may
-#   effectively render the measurement endless.
+energy_ramp = LinearEnergyRamp
 start_energy = 10
 delta_energy = 1
 end_energy = 100
 step_profile = "abrupt"
-endless = False
-constant_energy = False
 
 [devices]
 # Contains the configuration files of connected controllers and cameras.
--- a/src/viperleed/gui/measure/_defaults/_time_resolved.ini
+++ b/src/viperleed/gui/measure/_defaults/_time_resolved.ini
@@ -24,7 +24,7 @@ measurement_interval = 400
 [energies]
 # Settings of the energy ramp of the measurement
 # ----------------------------------------------
-# energy_ramp
+# ramp_type
 #   The energy ramp type the measurement will follow.
 # start_energy
 #   The energy at which the time-resolved measurement will start.
@@ -41,7 +41,7 @@ measurement_interval = 400
 #   means current energy). Also, fraction = 1 should not be explicitly
 #   included as the last step (nor should its settle time), as this is
 #   added automatically.
-energy_ramp = LinearEnergyRamp
+ramp_type = LinearEnergyRamp
 start_energy = 10
 delta_energy = 1
 end_energy = 100
--- a/src/viperleed/gui/measure/_defaults/_time_resolved.ini
+++ b/src/viperleed/gui/measure/_defaults/_time_resolved.ini
@@ -41,7 +41,7 @@ measurement_interval = 400
 #   means current energy). Also, fraction = 1 should not be explicitly
 #   included as the last step (nor should its settle time), as this is
 #   added automatically.
-ramp_type = LinearEnergyRamp
+ramp_type = ConstantEnergyRamp
 start_energy = 10
 delta_energy = 1
 end_energy = 100
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -1,7 +1,7 @@
 """Module energyramp of viperleed.gui.measure.classes.
 
 This module defines the EnergyRampABC class and various subclasses,
-which decides which energy is next during a measurement.
+which decide which energy is next during a measurement.
 """
 
 __authors__ = (
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -19,7 +19,8 @@ from viperleed.gui.measure.classes.abc import QMetaABC
 from viperleed.gui.measure.classes.abc import QObjectSettingsErrors
 from viperleed.gui.measure.classes.abc import QObjectWithError
 from viperleed.gui.measure.classes.settings import NotASequenceError
-from viperleed.gui.measure.dialogs.settingsdialog import SettingsDialogOption, SettingsTag
+from viperleed.gui.measure.dialogs.settingsdialog import SettingsDialogOption
+from viperleed.gui.measure.dialogs.settingsdialog import SettingsTag
 from viperleed.gui.measure.widgets.spinboxes import CoercingDoubleSpinBox
 
 
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -36,7 +36,7 @@ MINIMUM_ENERGY = 0.0
 MINIMUM_DELTA = 1e-4
 
 
-def get_matching_energy_ramp(settings):
+def get_ramp_from_settings(settings):
     """Determine and return the type of energy ramp."""
     constant_energy = False
     endless = False
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -183,7 +183,7 @@ class EnergyRampABC(QObjectWithError):                                          
             # Not a float
             self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
                             'energies/start_energy', '')
-            self._start_energy = max(self.min_energy, DEFAULT_START)
+            start_e = DEFAULT_START
         self._start_energy = max(self.min_energy, start_e)
         try:
             profile = settings.getsequence('energies', 'step_profile',
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -456,8 +456,7 @@ class LinearEnergyRamp(EnergyRampABC):
                                      tags=SettingsTag.REGULAR)
         end_widget = CoercingDoubleSpinBox(decimals=1, soft_range=(0, 1000),
                                            suffix=' eV')
-        tip = ('<nobr>The energy value at which </nobr>'
-               'the measurement will finish.')
+        tip = '<nobr>The energy at which the measurement ends.</nobr>'
         end = SettingsDialogOption('end_energy', end_widget, tooltip=tip,
                                    display_name=END_E_NAME,
                                    tags=SettingsTag.REGULAR)
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -40,6 +40,9 @@ MINIMUM_DELTA = 1e-4
 class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      # TODO: Move profile settings over to controller settings.
     """Generic energy ramp class."""
 
+    display_name = None
+    info_text = None
+
     def __init__(self, *args, **kwargs):
         """Initialise generic energy ramp class."""
         super().__init__(*args, **kwargs)
@@ -407,6 +410,10 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      
 class LinearEnergyRamp(EnergyRampABC):
     """Generic linear energy ramp."""
 
+    display_name = 'Linear energy ramp'
+    info_text = ('<nobr>Linearly increases or decreases the energy'
+                 '</nobr> until reaching the end energy.')
+
     def __init__(self, *args, **kwargs):
         """Initialize LinearEnergyRamp."""
         self._delta_energy = DEFAULT_DELTA
@@ -555,6 +562,11 @@ class LinearEnergyRamp(EnergyRampABC):
 class ConstantEnergyRamp(EnergyRampABC):
     """Constant energy ramp."""
 
+    display_name = 'Constant energy ramp'
+    info_text = ('<nobr>Keeps the energy constant throughout the</nobr>'
+                 ' measurement. Useful for time-resolved measurements '
+                 'at fixed energy.')
+
     @property
     def energy_steps(self):
         """Return the number of energy steps."""
@@ -582,6 +594,11 @@ class ConstantEnergyRamp(EnergyRampABC):
 
 class EndlessLinearEnergyRamp(LinearEnergyRamp):
     """Endless linear energy ramp."""
+
+    display_name = 'Endless linear energy ramp'
+    info_text = ('<nobr>Linearly increases or decreases the energy'
+                 '</nobr> until reaching the end energy. Energy resets '
+                 'to start when reaching the end.')
 
     @property
     def energy_steps(self):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -51,7 +51,7 @@ def get_matching_energy_ramp(settings):
     if constant_energy:
         return ConstantEnergyRamp
     if endless:
-        return EndlessLinearEnergyRamp
+        return SawtoothEnergyRamp
     return LinearEnergyRamp
 
 
@@ -593,10 +593,10 @@ class ConstantEnergyRamp(EnergyRampABC):
         return False  # A constant energy ramp is never finished.
 
 
-class EndlessLinearEnergyRamp(LinearEnergyRamp):
-    """Endless linear energy ramp."""
+class SawtoothEnergyRamp(LinearEnergyRamp):
+    """Sawtooth energy ramp."""
 
-    display_name = 'Endless linear energy ramp'
+    display_name = 'Sawtooth energy ramp'
     info_text = ('<nobr>Linearly increases or decreases the energy'
                  f'</nobr> until reaching the {END_E_NAME}. Energy resets '
                  f'to {START_E_NAME} when reaching the {END_E_NAME}.')
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -569,8 +569,9 @@ class ConstantEnergyRamp(EnergyRampABC):
         return True, ''
 
     def increment_energy(self):
-        """Go to the next energy in the ramp."""
-        # This is a no-op as the energy should never change.
+        """Advance the ramp state without changing the energy."""
+        # This is necessary to avoid recalling set_leed_energy.
+        self._previous_energy = self.current_energy
 
     def ramp_finished(self):                                                    # TODO: we could add a counter here that allows us to repeat constant energy ramps only a limited amount of times
         """Return whether the energy ramp has been finished."""
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -105,7 +105,7 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      
         The returned value excludes the very last step, i.e.,
         self.current_energy and the settling time for it.
         A typical call to set_leed_energy would be
-            self.set_leed_energy(*self.step_profile,
+            self.set_leed_energy(*self._energy_ramp.step_profile,
                                  self.current_energy,
                                  last_settle_time,
                                  ...)
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -38,7 +38,7 @@ class EnergyRampABC(QObjectWithError):                                          
         self._min_energy = MINIMUM_ENERGY
         self._previous_energy = 0.0
         self._start_energy = DEFAULT_START
-        self._step_profile = tuple()
+        self._step_profile = (ABRUPT,)
 
     @property
     @abstractmethod
@@ -168,7 +168,7 @@ class EnergyRampABC(QObjectWithError):                                          
         if not settings:
             self._min_energy = MINIMUM_ENERGY
             self._start_energy = max(self.min_energy, DEFAULT_START)            # TODO: without settings this will default to 0.0 right now. (Used to be 5.0 for energy calibration)
-            self._step_profile = ()
+            self._step_profile = (ABRUPT,)
             return
         try:
             self._min_energy = settings.getfloat('energies', 'min_energy',
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -607,14 +607,9 @@ class SawtoothEnergyRamp(LinearEnergyRamp):
 
     def increment_energy(self):
         """Go to the next energy in the ramp."""
-        if self._delta_energy > 0:
-            if self.current_energy + self._delta_energy > self._end_energy:
-                self.current_energy = self.start_energy
-                return
-        if self._delta_energy < 0:
-            if self.current_energy + self._delta_energy < self._end_energy:
-                self.current_energy = self.start_energy
-                return
+        if super().ramp_finished():
+            self.current_energy = self.start_energy
+            return
         super().increment_energy()
 
     def ramp_finished(self):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -501,12 +501,11 @@ class LinearEnergyRamp(EnergyRampABC):
 
     def ramp_finished(self):
         """Return whether the energy ramp has been finished."""
-        if self._delta_energy > 0:
-            if self.current_energy + self._delta_energy > self._end_energy:
-                return True
-        if self._delta_energy < 0:
-            if self.current_energy + self._delta_energy < self._end_energy:
-                return True
+        next_energy = self.current_energy + self._delta_energy
+        if self._delta_energy > 0 and next_energy > self._end_energy:
+            return True
+        if self._delta_energy < 0 and next_energy < self._end_energy:
+            return True
         return False
 
     def set_ramp(self, settings):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -39,6 +39,8 @@ MINIMUM_DELTA = 1e-4
 
 def get_matching_energy_ramp(settings):
     """Determine and return the type of energy ramp."""
+    constant_energy = False
+    endless = False
     try:
         constant_energy = settings.getboolean('energies', 'constant_energy',
                                               fallback=False)
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -655,5 +655,8 @@ class SawtoothEnergyRamp(LinearEnergyRamp):
         return False  # An endless energy ramp is never finished.
 
 
-ALL_ENERGY_RAMPS = (LinearEnergyRamp, SawtoothEnergyRamp,
-                    ConstantEnergyRamp)
+ALL_ENERGY_RAMPS = (
+    ConstantEnergyRamp,
+    LinearEnergyRamp,
+    SawtoothEnergyRamp,
+    )
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -199,8 +199,8 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):
         """
         return True, ''
 
-    @classmethod
-    def are_settings_invalid(cls, settings):
+    @staticmethod
+    def are_settings_invalid(settings):
         """Check if there are any invalid settings.
 
         Parameters
@@ -211,13 +211,10 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):
         Returns
         -------
         invalid_settings : list of tuples
-            Invalid _mandatory_settings of cls as a list of tuples.
-            The first entry in each tuple can be either '<section>',
-            '<section>/<option>', or
+            Invalid _mandatory_settings of ramp_class as a list of
+            tuples. The first entry in each tuple can be either
+            '<section>', '<section>/<option>', or
             '<section>/<option> not one of <value1>, <value2>, ...'.
-            Further optional entries may be added by subclasses. They
-            specify additional information on what is wrong with each
-            invalid setting.
         """
         ramp_name = settings.get('energies', 'ramp_type', fallback=None)
         for ramp in ALL_ENERGY_RAMPS:
@@ -226,6 +223,7 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):
                 break
         else:
             ramp_class = get_ramp_from_settings(settings)
+        # pylint: disable-next=protected-access
         mandatory_settings = ramp_class._mandatory_settings
         invalid_settings = settings.misses_settings(*mandatory_settings)
         return [(invalid,) for invalid in invalid_settings]
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -232,27 +232,12 @@ class EnergyRampABC(QObjectWithError):                                          
         -------
         None.
         """
-        # Validate overall structure: need an even number of entries and at
-        # least one (fraction, time) pair.
         try:
             n_items = len(profile)
         except TypeError:
-            self.emit_error(
-                QObjectSettingsErrors.INVALID_SETTINGS,
-                'energies/step_profile',
-                'Custom step profile must be a finite sequence of '
-                '(fraction, time) pairs'
-            )
-            self._step_profile = (ABRUPT,)
-            return
-
-        if n_items < 2 or n_items % 2 != 0:
-            self.emit_error(
-                QObjectSettingsErrors.INVALID_SETTINGS,
-                'energies/step_profile',
-                'Invalid custom step profile: expected an even number of '
-                f'entries (fraction, time, ...), found {n_items}'
-            )
+            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                            'energies/step_profile',
+                            'Step profiles must be sequences.')
             self._step_profile = (ABRUPT,)
             return
 
@@ -264,9 +249,19 @@ class EnergyRampABC(QObjectWithError):                                          
             if time_ < 0:
                 self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
                                 'energies/step_profile',
-                                '\nInfo: Time intervals must be non-negative')
+                                'Time intervals must be non-negative.')
                 self._step_profile = (ABRUPT,)
                 return
+
+        if n_items < 2 or n_items % 2 != 0:
+            self.emit_error(
+                QObjectSettingsErrors.INVALID_SETTINGS,
+                'energies/step_profile',
+                'Invalid custom step profile: expected an even number of '
+                f'entries (fraction, time, ...), found {n_items}.'
+            )
+            self._step_profile = (ABRUPT,)
+            return
         self._step_profile = profile
 
     def _set_linear_profile(self, profile):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -413,7 +413,7 @@ class LinearEnergyRamp(EnergyRampABC):
 
     display_name = 'Linear energy ramp'
     info_text = ('<nobr>Linearly increases or decreases the energy'
-                 f' </nobr>until reaching the {END_E_NAME}.')
+                 f' </nobr>until reaching {END_E_NAME}.')
 
     def __init__(self, *args, **kwargs):
         """Initialize LinearEnergyRamp."""
@@ -590,8 +590,8 @@ class SawtoothEnergyRamp(LinearEnergyRamp):
 
     display_name = 'Sawtooth energy ramp'
     info_text = ('<nobr>Linearly increases or decreases the energy'
-                 f' </nobr>until reaching the {END_E_NAME}. Energy resets '
-                 f'to {START_E_NAME} when reaching the {END_E_NAME}.')
+                 f' </nobr>until reaching {END_E_NAME}. Energy resets '
+                 f'to {START_E_NAME} when reaching {END_E_NAME}.')
 
     @property
     def n_steps(self):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -413,7 +413,7 @@ class LinearEnergyRamp(EnergyRampABC):
 
     display_name = 'Linear energy ramp'
     info_text = ('<nobr>Linearly increases or decreases the energy'
-                 f'</nobr> until reaching the {END_E_NAME}.')
+                 f' </nobr>until reaching the {END_E_NAME}.')
 
     def __init__(self, *args, **kwargs):
         """Initialize LinearEnergyRamp."""
@@ -590,7 +590,7 @@ class SawtoothEnergyRamp(LinearEnergyRamp):
 
     display_name = 'Sawtooth energy ramp'
     info_text = ('<nobr>Linearly increases or decreases the energy'
-                 f'</nobr> until reaching the {END_E_NAME}. Energy resets '
+                 f' </nobr>until reaching the {END_E_NAME}. Energy resets '
                  f'to {START_E_NAME} when reaching the {END_E_NAME}.')
 
     @property
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -293,17 +293,29 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):
             self._step_profile = (ABRUPT,)
             return
 
-        # Casting may break. This is caught in self._set_step_profile.
-        for fraction in profile[::2]:
-            float(fraction)
-        for time_ in profile[1::2]:
-            time_ = int(time_)
-            if time_ < 0:
-                self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
-                                'energies/step_profile',
-                                'Time intervals must be non-negative.')
-                self._step_profile = (ABRUPT,)
-                return
+        # If the first element is not a number, it's probably a named profile.
+        # We let the exception propagate to self._set_step_profile.
+        if n_items > 0:
+            float(profile[0])
+
+        try:
+            for fraction in profile[::2]:
+                float(fraction)
+            for time_ in profile[1::2]:
+                time_ = int(time_)
+                if time_ < 0:
+                    self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                                    'energies/step_profile',
+                                    'Time intervals must be non-negative.')
+                    self._step_profile = (ABRUPT,)
+                    return
+        except (ValueError, TypeError):
+            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                            'energies/step_profile',
+                            'Invalid custom step profile: fractions must be '
+                            'floats and times must be integers.')
+            self._step_profile = (ABRUPT,)
+            return
 
         if n_items < 2 or n_items % 2 != 0:
             self.emit_error(
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -418,6 +418,8 @@ class LinearEnergyRamp(EnergyRampABC):
     def __init__(self, *args, **kwargs):
         """Initialize LinearEnergyRamp."""
         self._delta_energy = DEFAULT_DELTA
+        # _end_energy is limited below by a minimum energy (as found
+        # in 'energies/min_energy' if present, 0.0 eV otherwise).
         self._end_energy = DEFAULT_END
         super().__init__(*args, **kwargs)
 
@@ -548,6 +550,7 @@ class LinearEnergyRamp(EnergyRampABC):
                 QObjectSettingsErrors.INVALID_SETTING_WITH_FALLBACK,
                 '', 'energies/end_energy', DEFAULT_END
                 )
+        self._end_energy = max(self.min_energy, self._end_energy)
 
 
 class ConstantEnergyRamp(EnergyRampABC):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -543,7 +543,7 @@ class LinearEnergyRamp(EnergyRampABC):
         if abs(self._delta_energy) < MINIMUM_DELTA:
             self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
                             'energies/delta_energy', f'{DELTA_E_NAME} was '
-                            'set to 0. Use constant energy mode instead.')
+                            'set to 0. Use constant-energy mode instead.')
         try:
             self._end_energy = settings.getfloat('energies', 'end_energy')
         except (TypeError, ValueError):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -383,7 +383,7 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      
             profile = (ABRUPT,)
         self._step_profile = profile
 
-    def _step_profile_from_strings(self, profile):                              # TODO: Warn for .ini files created before 23/05/2025.
+    def _step_profile_from_strings(self, profile):
         """Return a tuple of energies and times from strings."""
         delta = self.current_energy - self.previous_energy
         if abs(delta) < MINIMUM_DELTA:
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -14,6 +14,7 @@ __license__ = 'GPLv3+'
 
 from abc import abstractmethod
 from configparser import NoSectionError, NoOptionError
+import math
 
 from viperleed.gui.measure.classes.abc import QMetaABC
 from viperleed.gui.measure.classes.abc import QObjectSettingsErrors
@@ -52,6 +53,7 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      
     @abstractmethod
     def energy_steps(self):
         """Return the number of energy steps."""
+        return 0
 
     @property
     def current_energy(self):
@@ -421,7 +423,7 @@ class LinearEnergyRamp(EnergyRampABC):
         """Return the number of energy steps."""
         if self._delta_energy == 0.0:
             return 0
-        return 1 + round(self.energy_range/abs(self._delta_energy))
+        return 1 + math.floor(self.energy_range/abs(self._delta_energy))
 
     @classmethod
     def get_settings_widgets(cls):
--- a/src/viperleed/gui/measure/classes/energyramp.py
+++ b/src/viperleed/gui/measure/classes/energyramp.py
@@ -423,7 +423,7 @@ class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):
 class LinearEnergyRamp(EnergyRampABC):
     """Generic linear energy ramp."""
 
-    display_name = 'Linear energy ramp'
+    display_name = 'Linear'
     info_text = ('<nobr>Linearly increases or decreases the energy'
                  f' </nobr>up to and including {END_E_NAME}.')
 
@@ -567,7 +567,7 @@ class LinearEnergyRamp(EnergyRampABC):
 class ConstantEnergyRamp(EnergyRampABC):
     """Constant energy ramp."""
 
-    display_name = 'Constant energy ramp'
+    display_name = 'Constant'
     info_text = ('<nobr>Keeps the energy constant throughout the</nobr>'
                  ' measurement. Useful for time-resolved measurements '
                  'at fixed energy.')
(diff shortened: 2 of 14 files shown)


src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,465 @@
...
+        for time_ in profile[1::2]:
+            time_ = int(time_)
+            if time_ < 0:
+                self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                                'energies/step_profile',
+                                '\nInfo: Time intervals must be non-negative')
+                self._step_profile = (ABRUPT,)
+                return
+        self._step_profile = profile
+
```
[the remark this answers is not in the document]
FlorianDoerr: While the check suggested in this comment is sensible, one has to be careful with how it is implemented. The implementation by Copilot is of course broken and I'll have to edit it.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,465 @@
...
+            # Not present
+            self._delta_energy = DEFAULT_DELTA
+            self.emit_error(
+                QObjectSettingsErrors.INVALID_SETTING_WITH_FALLBACK,
+                '', 'energies/delta_energy', DEFAULT_DELTA
+                )
+        if self._delta_energy == 0.0:
+            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                            'energies/delta_energy', 'delta_energy was '
+                            'set to 0. Use constant energy mode instead.')
```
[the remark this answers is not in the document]
FlorianDoerr: Error emission of course aborts the measurement.

src/viperleed/gui/measure/measurement/time_resolved.py
```diff
@@ -201,8 +130,8 @@ def energy_step_duration(self, new_interval):
         """Set the duration of one energy step (msec)."""
         min_t = self.energy_step_duration_min
         if new_interval < min_t:
-            raise ValueError(f"{new_interval} is too small for the "
-                             f"controllers. Minimum is {min_t}")
+            raise ValueError(f'{new_interval} is too small for the '
+                             f'controllers. Minimum is {min_t}')
         self.settings.set('measurement_settings', 'measurement_interval',
```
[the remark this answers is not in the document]
FlorianDoerr: Looks like we never used this property but this is indeed a rough bug.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,484 @@
...
+            pass
+        try:
+            endless = settings.getboolean('energies', 'endless')
+        except (TypeError, ValueError, NoSectionError, NoOptionError):
+            pass
+        if constant_energy:
+            return ConstantEnergyRamp
+        if endless:
+            return EndlessLinearEnergyRamp
+        return LinearEnergyRamp
```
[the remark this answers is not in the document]
FlorianDoerr: @copilot apply changes based on [this feedback](https://github.com/viperleed/viperleed/pull/519#discussion_r2948106128)

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,480 @@
...
+        else:
+            return
+
+        shape, *_ = profile
+        if shape.lower() == ABRUPT:
+            pass
+        elif shape.lower() == LINEAR:
+            self._set_linear_profile(profile)
+            return
+        else:
```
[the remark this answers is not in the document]
FlorianDoerr: This is wrong.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,601 @@
...
+
+from abc import abstractmethod
+from configparser import NoSectionError, NoOptionError
+
+from viperleed.gui.measure.classes.abc import QMetaABC
+from viperleed.gui.measure.classes.abc import QObjectSettingsErrors
+from viperleed.gui.measure.classes.abc import QObjectWithError
+from viperleed.gui.measure.classes.settings import NotASequenceError
+from viperleed.gui.measure.dialogs.settingsdialog import SettingsDialogOption, SettingsTag
+from viperleed.gui.measure.widgets.spinboxes import CoercingDoubleSpinBox
```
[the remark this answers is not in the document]
FlorianDoerr: No idea how Copilot managed to be both wrong and right at the same time, but obviously the line that is too long is line 22 and not line 23. Just figured this was funny.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,602 @@
...
+        """Return whether the energy ramp has been finished."""
+        return False  # A constant energy ramp is never finished.
+
+
+class EndlessLinearEnergyRamp(LinearEnergyRamp):
+    """Endless linear energy ramp."""
+
+    @property
+    def energy_steps(self):
+        """Return the number of energy steps."""
```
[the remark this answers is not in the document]
FlorianDoerr: This is wrong, energy_steps works as intended in the described case.

src/viperleed/gui/measure/measurement/energy_calibration.py
```diff
@@ -165,27 +93,25 @@ def are_runtime_settings_ok(self):
...
+        if self._energy_ramp.energy_range < 10:
             # Require at least 10 eV for a reasonable calibration
             self.emit_error(
                 QObjectSettingsErrors.INVALID_SETTINGS,
                 'energies/start_energy and /end_energy',
-                f"\nToo small energy range ({abs(egy_range)} eV) for "
-                "calibration. It should be at least 10 eV."
+                '\nToo small energy range '
+                f'({self._energy_ramp.energy_range} eV) '
+                'for calibration. It should be at least 10 eV.'
```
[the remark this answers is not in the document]
FlorianDoerr: This can no longer happen during energy calibration. (Because we only allow linear energy ramps.)

src/viperleed/gui/measure/_defaults/_time_resolved.ini
```diff
@@ -24,6 +24,8 @@ measurement_interval = 400
 [energies]
 # Settings of the energy ramp of the measurement
 # ----------------------------------------------
+# ramp_type
```
michele-riva, line 27: Considering that this default file is for a triggered time-resolved measurement, and that these measurement types would normally be used primarily for I(t) measurements, it may be more fitting to have a ConstantEnergyRamp as the default.

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -37,17 +40,21 @@
...
 MAX_NUM_STEPS = 7
 MAX_DELAY = 65535
 N_COLUMNS = 2
 N_HEADER_ROWS = 2
+ALLOWED_ENERGY_RAMPS = {
+    'IVVideo': (LinearEnergyRamp,),
+    'MeasureEnergyCalibration': (LinearEnergyRamp,),
+    'TimeResolved': (LinearEnergyRamp, EndlessLinearEnergyRamp,
+                     ConstantEnergyRamp),
+    }
```
michele-riva, line 56: Perhaps it would be better to only list here the classes that support a subset of the ramps. This would reduce a bit the maintenance burden.

A simple way would be to have an ALL_ENERGY_RAMPS global. Then, when populating the combo box, do `ALLOWED_ENERGY_RAMPS.get(meas_cls, ALL_ENERGY_RAMPS)` rather than using `__getitem__`

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +197,117 @@ def update_widgets(self):
...
+        self._settings = settings
+        kwargs.setdefault('display_name', 'Energy Ramp')
+        kwargs.setdefault('tags', SettingsTag.REGULAR)
+        kwargs.setdefault('tooltip', 'This section handles settings that '
+                          'control how the energy is ramped between steps.')
+        super().__init__(**kwargs)
+        self._step_profile = StepProfileViewer()
+        self._ramp_type = qtw.QComboBox()
+        self._energy_options = tuple()
+        self.settings_changed.connect(self._store_energy_ramp_settings)
```
michele-riva: Perhaps this connection should go to _compose_and_connect?

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +197,117 @@ def update_widgets(self):
...
+
+    def _compose_and_connect(self):
+        """Compose widgets and connect relevant signals."""
+        layout = qtw.QFormLayout()
+        tip = 'The shape of the energy ramp.'
+        label = InfoLabel(label_text='Ramp type', tooltip=tip)
+        layout.addRow(label, self._ramp_type)
+        meas_cls = self._settings["measurement_settings"]["measurement_class"]
+        for ramp_cls in ALLOWED_ENERGY_RAMPS[meas_cls]:
+            self._ramp_type.addItem(ramp_cls.__name__, userData=ramp_cls)
```
michele-riva: We may want to have some more user-friendly names displayed in the combo box. Consider adding a display_name class attribute to EnergyRampABC

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +197,117 @@ def update_widgets(self):
...
+        while self.central_widget.layout().rowCount() > 2:
+            self.central_widget.layout().removeRow(2)
+        selected_ramp = self._ramp_type.currentData()
+        self._energy_options = selected_ramp.get_settings_widgets()
+        if self._settings.has_option('energies', 'min_energy'):
+            min_energy = self._settings.getfloat('energies', 'min_energy')
+            start = next(option for option in self._energy_options
+                         if option.option_name == 'start_energy')
+            with qtc.QSignalBlocker(start.handler_widget):
+                start.handler_widget.soft_minimum = min_energy
```
michele-riva: This is only there for the energy calibration, I guess? Is there a nicer way to do this (e.g., only done in the energy calibration get_settings_handler) than having this in here?

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +197,117 @@ def update_widgets(self):
...
+
+    @qtc.pyqtSlot()
+    def _store_energy_ramp_settings(self):
+        """Store the selected energy-ramp settings."""
+        self._settings.set('energies', 'ramp_type',
+                           self._ramp_type.currentData().__name__)
+        self._settings.set('energies', 'step_profile',
+                           self._step_profile.get_())
+        for option in self._energy_options:
+            self._settings.set('energies', option.option_name, option.get_())
```
michele-riva, line 315: This may leave behind spurious settings that do not apply to the current ramp type. Is this intended?

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +197,117 @@ def update_widgets(self):
...
+
+    @qtc.pyqtSlot()
+    def update_widgets(self):
+        """Update widgets from settings."""
+        with qtc.QSignalBlocker(self):
```
michele-riva, line 323: Perhaps good to comment on why we block signals here

src/viperleed/gui/measure/measurement/energy_calibration.py
```diff
@@ -45,83 +44,12 @@ class MeasureEnergyCalibration(MeasurementABC):
...
-    def _min_energy(self):
-        """Return the minimum starting energy (in eV)."""
-        try:
-            min_e = self.settings.getfloat('energies', 'min_energy',
-                                           fallback=5.)
-        except (TypeError, ValueError):
-            # Not a float
-            min_e = 5.0
-        return min_e
+        self._old_coefficients = ''
```
michele-riva, line 46: Perhaps if we would create self._old_coefficients before the call to super we could simplify abort further by removing the hasattr check

src/viperleed/gui/measure/measurement/time_resolved.py
```diff
@@ -331,11 +259,11 @@ def are_runtime_settings_ok(self):
...
             return False
 
-        num_meas = (1 + round((self._end_energy - self.start_energy)
-                              / self._delta_energy))
         if self.is_continuous:
             self._prepare_continuous_mode()
-            self.data_points.nr_steps_total = num_meas                          # TODO: incorrect for endless and constant energy. Also, unused now that we use different DataPoints for plotting and measurement.
+            self.data_points.nr_steps_total = max(
+                1, self._energy_ramp.energy_steps
+                )
```
michele-riva: Is the comment about this value being unused because we use a copy still relevant? If yes, then perhaps we should leave it on (or fix it).

src/viperleed/gui/measure/measurement/_meassettings.py
michele-riva: Perhaps, it may also be useful to have either a tooltip for the items or a short description below the combo (which updates based on the selection). Consider reusing the old tooltips that were associated with the constant_energy and endless tick boxes.

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -319,66 +322,6 @@ def set_secondary_controllers(self, new_controllers):
...
-            return self.settings.getfloat('energies', 'start_energy',
-                                          fallback=0)
-        except (TypeError, ValueError):
-            # Not a float
-            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
-                            'energies/start_energy', '')
-            return 0.0
-
-    @property
-    def step_profile(self):                                                     # TODO: probably move to generator class?
```
michele-riva, line 337: You could have retained this property to simply access the _energy_ramp ones'. Then there would be perhaps fewer changes in the other classes.

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -778,6 +699,22 @@ def set_settings(self, new_settings):
                               if c.measures()}
         for camera in self.cameras:
             self._missing_data[camera] = 1
+
+        ramp_name = self.settings.get('energies', 'ramp_type', fallback=None)
```
michele-riva: perhaps this part of code could go to a dedicated private method, e.g., `_prepare_energy_ramp()`?

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -778,6 +699,22 @@ def set_settings(self, new_settings):
...
+        ramp_type = None
+        if ramp_name:
+            try:
+                ramp_type = base.class_from_name('classes', ramp_name)
+            except (RuntimeError, ValueError):
+                pass
+        if ramp_type and not issubclass(ramp_type, EnergyRampABC):
+            ramp_type = None
```
michele-riva: May want to notify this to the users with an error_occured. Same if class_from_name fails to pull ramp_name (likely only ValueError)

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -1181,69 +1121,20 @@ def _get_device_settings(self, configname):
...
-        # f[k] = (t[k] + t[k+1]) / (2*tot_time)
-        slope = delta_e / (2 * tot_time)
-        energies = (slope*(ti + tj) + self.current_energy
-                    for ti, tj in zip(times, times[1:]))
-
-        # Finally interleave energies and times
-        return tuple(v for tup in zip(energies, intervals) for v in tup)
-
-    @abstractmethod
     def _is_finished(self):
```
michele-riva: Perhaps we should consider a small revision of the use of this _is_finished. It may make sense to remove it entirely. I've never really liked the side effect of incrementing the energy inside it if the measurement is not finished, as it is not clear from the name.
It may be possible to get rid of it entirely (perhaps adding a clearer interface if any measurement needs to do anything when it is really finished, e.g., `_on_energy_ramp_finished`) now that it is not abstract and has been removed from other subclasses.

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -1119,6 +1056,9 @@ def _finalize(self, *_):
 
             # Keep only the primary controller connected, so we can
             # set the LEED energy to zero (and detect it has been set)
+            if self.current_energy == 0:
+                self._cleanup_and_end()
+                return
```
michele-riva, line 1046: Probably better to move these lines before the previous comment, that refers to the next bit of code.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+START_E_NAME = 'E start'
+END_E_NAME = 'E end'
+DEFAULT_DELTA = 0.5
+DEFAULT_END = 0.0
+DEFAULT_START = 0.0
+MINIMUM_ENERGY = 0.0
+MINIMUM_DELTA = 1e-4
+
+
+class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      # TODO: Move profile settings over to controller settings.
```
michele-riva: About the TODO: we should discuss if we should really do this. I am conflicted. While it makes sense conceptually, it may simply become annoying much like the settle times, as one probably wants to use different profiles for different measurements (e.g., e cal vs. iv)

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        super().__init__(*args, **kwargs)
+        self._current_energy = 0.0
+        self._min_energy = MINIMUM_ENERGY
+        self._previous_energy = 0.0
+        self._start_energy = DEFAULT_START
+        self._step_profile = (ABRUPT,)
+
+    @property
+    @abstractmethod
+    def energy_steps(self):
```
michele-riva: Perhaps n_steps is clearer? When I read this I could misinterpret it as returning a sequence of the energies.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+
+class EnergyRampABC(QObjectWithError, metaclass=QMetaABC):                      # TODO: Move profile settings over to controller settings.
+    """Generic energy ramp class."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialise generic energy ramp class."""
+        super().__init__(*args, **kwargs)
+        self._current_energy = 0.0
+        self._min_energy = MINIMUM_ENERGY
+        self._previous_energy = 0.0
```
michele-riva, line 71: Perhaps we should set this to None? What if a unit is at some energy (for whatever reason, e.g., another measurement failed to finish cleanly) and we think we are already at the previous energy? Would we bypass setting the energy altogether?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+from viperleed.gui.measure.classes.abc import QObjectSettingsErrors
+from viperleed.gui.measure.classes.abc import QObjectWithError
+from viperleed.gui.measure.classes.settings import NotASequenceError
+from viperleed.gui.measure.dialogs.settingsdialog import SettingsDialogOption
+from viperleed.gui.measure.dialogs.settingsdialog import SettingsTag
+from viperleed.gui.measure.widgets.spinboxes import CoercingDoubleSpinBox
+
+
+ABRUPT = 'abrupt'
+LINEAR = 'linear'
```
michele-riva, line 28: Some versions of these constants seem to appear in multiple places: here, _meas_settings, and in the names of the step profile editor widgets. Would be nice to unify them.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        return self._start_energy
+
+    @property
+    def step_profile(self):
+        """Return a list of energies and times for setting the next energy.
+
+        The returned value excludes the very last step, i.e.,
+        self.current_energy and the settling time for it.
+        A typical call to set_leed_energy would be
+            self.set_leed_energy(*self.step_profile,
```
michele-riva:
```suggestion
            measurement.set_leed_energy(*self.step_profile,
```

+ adjust indents

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        except (ValueError, TypeError):
+            pass
+
+        shape, *params = self._step_profile
+        is_abrupt = shape.lower() == ABRUPT
+        values = tuple() if is_abrupt else self._get_linear_step(*params)
+        return values
+
+    @classmethod
+    def get_matching_energy_ramp(cls, settings):
```
michele-riva: This is actually a static method as it does not use cls. It may even be better to make into a module-level function. This way, it does not look like EnergyRampABC needs to know about its subclasses.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        """Determine and return the type of energy ramp."""
+        constant_energy = False
+        endless = False
```
michele-riva: Perhaps these should just be fallbacks for the getboolean calls below?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+
+    @classmethod
+    @abstractmethod
+    def ramp_settings_ok(cls, energy_options):
+        """Return whether the settings for the ramp are ok.
+
+        Parameters
+        ----------
+        energy_options : tuple
+            The energy options to check.
```
michele-riva: More documentation as to what the items should be

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +197,117 @@ def update_widgets(self):
...
+            for option in self._energy_options:
+                value = self._settings['energies'].getfloat(option.option_name,
+                                                            fallback=None)
+                if value is not None:
+                    option.handler_widget.setValue(value)
+            step_profile = self._settings['energies'].get('step_profile',
+                                                          fallback=None)
+            if step_profile:
+                self._step_profile.set_(step_profile)
+
```
michele-riva, line 374:
```suggestion


```

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        -------
+        None.
+        """
+        try:
+            self._set_custom_profile(profile)
+        except (ValueError, TypeError):
+            # Not a custom profile.
+            pass
+        else:
+            return
```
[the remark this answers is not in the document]
michele-riva: I think this suggestion makes some sense. Perhaps you could look into this?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+                                                 fallback=MINIMUM_ENERGY)
+        except (TypeError, ValueError):
+            # Not a float
+            self._min_energy = MINIMUM_ENERGY
+        try:
+            start_e = settings.getfloat('energies', 'start_energy',
+                                        fallback=DEFAULT_START)
+        except (TypeError, ValueError):
+            # Not a float
+            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
```
michele-riva: Review again these errors. Most of them are actually INVALID_..._WITH_FALLBACK

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+
+        Returns
+        -------
+        start_energy : float
+            The first energy of the energy ramp.
+        """
+        return self._start_energy
+
+    @property
+    def step_profile(self):
```
michele-riva, line 125: We should consider whether all this mess with step profiles would not be handled more easily with some simple dedicated classes

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+
+    @property
+    def energy_range(self):
+        """Return the energy range."""
+        return abs(self._end_energy - self._start_energy)
+
+    @property
+    def energy_steps(self):
+        """Return the number of energy steps."""
+        if self._delta_energy == 0.0:
```
michele-riva: Perhaps you want to check abs(delta)<MIN_DELTA instad?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        reason : str
+            A descriptive string elaborating why the settings
+            are not acceptable.
+        """
+        energy_dict = {option.option_name: float(option.get_())
+                       for option in energy_options}
+        start_energy = energy_dict.get('start_energy', DEFAULT_START)
+        delta_energy = energy_dict.get('delta_energy', DEFAULT_DELTA)
+        end_energy = energy_dict.get('end_energy', DEFAULT_END)
+        if delta_energy == 0.0:
```
michele-riva: perhaps you want to check abs < MIN?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+            return False, ('Delta energy cannot be zero.'
```
michele-riva: You should consider using the constants for the energy names in these messages

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+                           ' Use constant energy ramp.')
+        if delta_energy > 0 and end_energy < start_energy:
+            return False, ('For positive delta energy, end energy '
+                           'should be greater than start energy.')
+        if delta_energy < 0 and end_energy > start_energy:
+            return False, ('For negative delta energy, end energy '
+                           'should be less than start energy.')
+        return True, ''
+
+    def increment_energy(self):
```
michele-riva, line 545: We should consider whether it would make sense to use a rule for incrementing the energy that is more stable against floating-point summation errors (doing start+i*delta is better then doing start + delta + delta + ... + delta). It may not be relevant for the typical deltas we use in LEED, though.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+
+    def ramp_finished(self):
+        """Return whether the energy ramp has been finished."""
+        if self._delta_energy > 0:
+            if self.current_energy + self._delta_energy > self._end_energy:
+                return True
+        if self._delta_energy < 0:
+            if self.current_energy + self._delta_energy < self._end_energy:
+                return True
+        return False
```
michele-riva:
```suggestion
        next_energy = self.current_energy + self._delta_energy
        if self._delta_energy > 0 and next_energy > self._end_energy:
            return True
        if self._delta_energy < 0 and next_energy < self._end_energy:
            return True
        return False
```

michele-riva: Also, consider whether we should use MIN_DELTA instead of zero

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                            'energies/delta_energy', '')
+        except (NoSectionError, NoOptionError):
+            # Not present
+            self._delta_energy = DEFAULT_DELTA
+            self.emit_error(
+                QObjectSettingsErrors.INVALID_SETTING_WITH_FALLBACK,
+                '', 'energies/delta_energy', DEFAULT_DELTA
+                )
+        if self._delta_energy == 0.0:
```
michele-riva: also here compare abs to MIN

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+            self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                            'energies/delta_energy', 'delta_energy was '
+                            'set to 0. Use constant energy mode instead.')
```
michele-riva:
```suggestion
                            'set to 0. Use constant-energy mode instead.')
```

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        """Go to the next energy in the ramp."""
+        if self._delta_energy > 0:
+            if self.current_energy + self._delta_energy > self._end_energy:
+                self.current_energy = self.start_energy
+                return
+        if self._delta_energy < 0:
+            if self.current_energy + self._delta_energy < self._end_energy:
+                self.current_energy = self.start_energy
+                return
+        super().increment_energy()
```
michele-riva:
```suggestion
        if super().ramp_finished():
            self.current_energy = self.start_energy
            return
        super().increment_energy()
```

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+        """Advance the ramp state without changing the energy."""
+        # This is necessary to avoid recalling set_leed_energy.
+        self._previous_energy = self.current_energy
+
+    def ramp_finished(self):                                                    # TODO: we could add a counter here that allows us to repeat constant energy ramps only a limited amount of times
+        """Return whether the energy ramp has been finished."""
+        return False  # A constant energy ramp is never finished.
+
+
+class EndlessLinearEnergyRamp(LinearEnergyRamp):
```
michele-riva: Perhaps we should rename this to `SawtoothEnergyRamp`?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+    def getfloat(self, section, option, fallback=None):
+        """Return configured float values or a fallback."""
+        value = self._float_values.get((section, option), fallback)
+        if isinstance(value, Exception):
+            raise value
+        return value
+
+    def getsequence(self, section, option, fallback=None):
+        """Return the configured sequence or a fallback."""
+        return self._sequence
```
michele-riva: Perhaps you could simplify the tests a bit using a ViPErLEEDSettings object in _make_settings. Then you would not need the _FakeRampSettings. You could even have _make_settings already create the [energies] section and pass on only its contents as kwargs, like so:

```python
def _make_settings(**kwargs):
    settings_dict = {'energies': kwargs} if kwargs else {}
    settings = ViPErLEEDSettings()
    settings.read_dict(settings_dict)
    return settings
```

Then you would not need to raise NoOptionError, just don't pass the option.

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+        linear_ramp.set_ramp(None)
+
+        assert linear_ramp.min_energy == MINIMUM_ENERGY
+        assert linear_ramp.start_energy == DEFAULT_START
+        assert linear_ramp._step_profile == (ABRUPT,)
+        assert linear_ramp._delta_energy == DEFAULT_DELTA
+        assert linear_ramp._end_energy == DEFAULT_END
+
+    def test_set_ramp_clamps_start_energy_to_minimum(self, linear_ramp):
+        """Check that start_energy is never set below min_energy."""
```
michele-riva: I just realized that one should rather clamp end_energy in case delta is < 0.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,605 @@
...
+    def current_energy(self, new_energy):
+        """Set a new value of the current_energy.
+
+        Parameters
+        ----------
+        new_energy : float
+            The new current energy
+        """
+        self._previous_energy = self.current_energy
+        self._current_energy = new_energy
```
michele-riva, line 96: For a linear ramp, what if new_energy is outside the start-end interval? Do we need to care?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+        linear_ramp._current_energy = 10.0
+        linear_ramp._step_profile = (LINEAR, '5', '10')
+
+        assert linear_ramp.step_profile == pytest.approx(
+            (1.0, 2, 3.0, 2, 5.0, 2, 7.0, 2, 9.0, 2)
+            )
+
+    def test_get_linear_step_returns_empty_when_interval_is_zero(self,
+                                                                 linear_ramp):
+        """Check that no profile is generated when integer interval is zero."""
```
michele-riva: This name and docstring feel a bit confusing. Maybe you could add an assertion that the _step_profile is abrupt?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+
+    def test_step_profile_falls_back_to_abrupt(self, linear_ramp):
+        """Check ABRUPT fallback branch when parsing as fractions fails."""
+        linear_ramp._previous_energy = 3.0
+        linear_ramp._current_energy = 6.0
+        linear_ramp._step_profile = (ABRUPT,)
+
+        assert linear_ramp.step_profile == ()
+
+    def test_step_profile_falls_back_to_linear_profile_generation(self,
```
michele-riva: this test name is a bit confusing: "falls back" seems out of place

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+
+    def test_get_linear_step_returns_empty_when_interval_is_zero(self,
+                                                                 linear_ramp):
+        """Check that no profile is generated when integer interval is zero."""
+        linear_ramp._previous_energy = 0.0
+        linear_ramp._current_energy = 10.0
+
+        assert linear_ramp._get_linear_step(20, 5) == ()
+
+    def test_get_linear_step_returns_empty_for_tiny_energy_delta(self,
```
michele-riva, line 348: This test seems to be testing the wrong thing, as the profile is never set to linear (i.e., it is the same path as the previous test)

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+    def test_set_step_profile_rejects_non_sequence_profile(self, linear_ramp):
+        """Check that non-sequence profiles are rejected as invalid."""
+        linear_ramp._set_step_profile(42)
+        assert linear_ramp._step_profile == (ABRUPT,)
+
+
+class TestRampSettingsWidgetsAndValidation:
+    """Tests for get_settings_widgets and ramp_settings_ok methods."""
+
+    _invalid_linear_settings = {
```
michele-riva, line 370: Perhaps add a comment with info about what the items in the tuple are

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+            -1.0,
+            11.0,
+            'For negative delta energy',
+            ),
+        }
+
+    def test_linear_settings_widgets_include_expected_options(self, mocker):
+        """Check linear widget setup without creating real Qt widgets."""
+        mocker.patch.object(energyramp_mod, 'CoercingDoubleSpinBox',
+                            _FakeSpinBox)
```
michele-riva: Rather than importing the module and using mocker.patch.object, you can also have the module name as a global in the test file and mocker.patch it's attribute, like so:

```python
MODULE = 'viperleed.gui.measure.classes.energyramp'
...
mocker.patch(f'{MODULE}.CoercingDoubleSpinBox', _FakeSpinBox)
```

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+        """Check that incrementing updates both current and previous values."""
+        linear_ramp.current_energy = 10.0
+        linear_ramp._delta_energy = 1.5
+
+        linear_ramp.increment_energy()
+
+        assert linear_ramp.previous_energy == 10.0
+        assert linear_ramp.current_energy == pytest.approx(11.5)
+
+# pylint: disable=protected-access
```
michele-riva:
```suggestion

# pylint: disable=protected-access
```

src/viperleed/gui/measure/classes/energyramp.py
michele-riva: Similarly, a ConstantEnergyRamp should perhaps not allow setting any energy other than its start one?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+class TestLinearEnergyRampFinished:
+    """Tests for LinearEnergyRamp.ramp_finished."""
+
+    def test_not_finished_positive_delta_within_range(self, linear_ramp):
```
michele-riva: Perhaps all these not finished test could be combined with one parametrization? Same for the finished ones

michele-riva: Also, perhaps you could rather use set_ramp instead of accessing all the private attributes?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+
+
+class TestEndlessLinearEnergyRampIncrementEnergy:
+    """Tests for EndlessLinearEnergyRamp.increment_energy."""
+
+    def test_energy_steps_is_zero(self, endless_ramp):
+        """Check endless-ramp energy_steps value."""
+        assert endless_ramp.energy_steps == 0
+
+    def test_wraps_to_start_on_positive_overflow(self, endless_ramp):
```
michele-riva: Could use set_ramp here too, as well as reducing the number of functions by parametrization

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+        """Check that an ABRUPT profile is stored correctly."""
+        linear_ramp._set_step_profile((ABRUPT,))
+        assert linear_ramp._step_profile == (ABRUPT,)
+
+    def test_linear_profile_valid(self, linear_ramp):
+        """Check that a valid LINEAR profile is stored correctly."""
+        linear_ramp._set_step_profile((LINEAR, '5', '100'))
+        assert linear_ramp._step_profile == (LINEAR, '5', '100')
+
+    def test_linear_profile_too_few_params(self, linear_ramp):
```
michele-riva: Perhaps all the tests in which the linear becomes an abrupt one could be merged into a single parameterized one

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,687 @@
...
+        """Check that a valid custom (numeric) profile is stored correctly."""
+        linear_ramp._set_step_profile(('0.5', '50'))
+        assert linear_ramp._step_profile == ('0.5', '50')
+
+    def test_custom_profile_multiple_pairs(self, linear_ramp):
+        """Check that a multi-pair custom profile is stored correctly."""
+        linear_ramp._set_step_profile(('0.3', '30', '0.7', '70'))
+        assert linear_ramp._step_profile == ('0.3', '30', '0.7', '70')
+
+    def test_custom_profile_odd_length(self, linear_ramp):
```
michele-riva: parametrize like for the linear fallbacks

michele-riva (changes requested):
Looks good, though I have the usual bunch of suggestions.

src/viperleed/gui/measure/measurement/time_resolved.py
FlorianDoerr: This is still true, however there is a comment about this in datapoints.py already. Essentially, nr_steps_total is only used when plotting separate steps, which is currently not done.

src/viperleed/gui/measure/measurement/abc.py
FlorianDoerr: It is only accessed 4 times in the code and I think it is good to see immediately that the step profile is in fact a property of the ramp and not the measurement.

src/viperleed/gui/measure/measurement/_meassettings.py
FlorianDoerr: We originally intended to only have min_energy for the energy calibration. This will however work for any other .ini that contains a min_energy. I think this is much better than looking for the measurement type as it allows us to reuse the code in the future. To see this min_energy for the other measurement types, users would need to set min_energy in their .ini and if they want to, they can now.

src/viperleed/gui/measure/measurement/_meassettings.py
FlorianDoerr: self._energy_options should only ever be populated with the widgets required by the ramp type.

src/viperleed/gui/measure/classes/energyramp.py
FlorianDoerr: [Wikipedia](https://en.wikipedia.org/wiki/Machine_epsilon) indicates that a rounding error of double precision variables is going to be ~2.22e-16. This is not only not going to be noticable, one would need a measurement that is significantly longer than a human lifetime for this rounding error to be visible at all.

src/viperleed/gui/measure/classes/energyramp.py
FlorianDoerr: If we implement a new step type, then we can do this to reduce complexity. Otherwise, I would prefer to not implement a new class-abstraction layer. We are already doing a lot of abstraction, and we run the risk of obfuscating what is actually going on if we shove another 50 layers in between.

src/viperleed/gui/measure/classes/energyramp.py
FlorianDoerr: An Arduino would need to be trapped at a non-zero DAC output before the measurement was initialized and then the measurement would need to break before setting start energy. While I agree that there are ways for an Arduino to break and remain at a wrong DAC output, I do not think this is the right point address this issue. If anything, one could make the measurement safer by catching any exception at the GUI level and attempting to (re-)connect to the last known primary controller and forcing a zero eV output.

src/viperleed/gui/measure/classes/energyramp.py
FlorianDoerr: This is handled by the increment_energy logic. If used like intended, both cases you outline will not be violated. However, if someone wants to force a current_energy, they can by setting it via the setter and this behaviour is necessary. For instance, we want to be able to set the energy to zero at the end of a measurement. This would not be possible if there were a min_energy and if we were to enforce the range between start and end or a constant start energy.

tests/gui/measure/classes/test_energyramp.py
FlorianDoerr: It does not need to be set to linear as _get_linear_step is the call that is performed for linear profiles. It does indeed test the same path, but with a different fail condition. In this case we are testing that we do not return a linear step profile for energy increments smaller than MINIMUM_DELTA.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,610 @@
...
+START_E_NAME = 'E start'
+END_E_NAME = 'E end'
+DEFAULT_DELTA = 0.5
+DEFAULT_END = 0.0
+DEFAULT_START = 0.0
+MINIMUM_ENERGY = 0.0
+MINIMUM_DELTA = 1e-4
+
+
+def get_matching_energy_ramp(settings):
```
michele-riva: Perhaps `get_ramp_from_settings` would be clearer? Another alternative, simply `from_settings`, to be used as

```python
import energyramp
...
energyramp.from_settings(settings)
```

May also be good to include in the name the fact that we return a class (or a type) and not an actual instance.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,610 @@
...
+        return self._start_energy
+
+    @property
+    def step_profile(self):
+        """Return a list of energies and times for setting the next energy.
+
+        The returned value excludes the very last step, i.e.,
+        self.current_energy and the settling time for it.
+        A typical call to set_leed_energy would be
+            measurement.set_leed_energy(*self._energy_ramp.step_profile,
```
michele-riva:
```suggestion
            measurement.set_leed_energy(*self.step_profile,
```

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,610 @@
...
+            energies_times[2*i+1] = time_
+        return tuple(energies_times)
+
+
+class LinearEnergyRamp(EnergyRampABC):
+    """Generic linear energy ramp."""
+
+    display_name = 'Linear energy ramp'
+    info_text = ('<nobr>Linearly increases or decreases the energy'
+                 f'</nobr> until reaching the {END_E_NAME}.')
```
michele-riva: Perhaps the space should be before the `</nobr>` so it does not look like the second line is indented by one space?

Adjust also others.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,610 @@
...
+
+    display_name = 'Constant energy ramp'
+    info_text = ('<nobr>Keeps the energy constant throughout the</nobr>'
+                 ' measurement. Useful for time-resolved measurements '
+                 'at fixed energy.')
+
+    @property
+    def n_steps(self):
+        """Return the number of energy steps."""
+        return 0
```
michele-riva, line 611: Shouldn't this be technically 1 step?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -0,0 +1,610 @@
...
+        return False  # A constant energy ramp is never finished.
+
+
+class SawtoothEnergyRamp(LinearEnergyRamp):
+    """Sawtooth energy ramp."""
+
+    display_name = 'Sawtooth energy ramp'
+    info_text = ('<nobr>Linearly increases or decreases the energy'
+                 f'</nobr> until reaching the {END_E_NAME}. Energy resets '
+                 f'to {START_E_NAME} when reaching the {END_E_NAME}.')
```
michele-riva:
```suggestion
                 f'</nobr> until reaching {END_E_NAME}. Energy resets '
                 f'to {START_E_NAME} when reaching {END_E_NAME}.')
```

Probably the text looks funny otherwise? Check the others too.

To be discussed: It may be worth to be clearer for users (also on the linear one): is the end energy one of the energies or not (alternatively, this info could go to the E end tooltip)?

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +201,132 @@ def update_widgets(self):
...
+            'display_name' : Displayed name of section.
+            'tags' : Tags associated with the section.
+            'tooltip' : Tooltip displayed with the section.
+
+        Returns
+        -------
+        None.
+        """
+        self._settings = settings
+        kwargs.setdefault('display_name', 'Energy Ramp')
```
michele-riva: I have some suggestions about the text in these widgets/tooltips. Let's discuss in person

src/viperleed/gui/measure/measurement/_meassettings.py
```diff
@@ -190,6 +201,132 @@ def update_widgets(self):
...
+            self.central_widget.layout().addRow(*option)
+            option.value_changed.connect(self.settings_changed)
+            option.value_changed.connect(self.settings_ok_changed)
+        self.update_widgets()
+
+    @qtc.pyqtSlot()
+    def _store_energy_ramp_settings(self):
+        """Store the selected energy-ramp settings."""
+        self._settings.set('energies', 'ramp_type',
+                   self._ramp_type.combo_box.currentData().__name__)
```
michele-riva: indent

src/viperleed/gui/measure/measurement/energy_calibration.py
```diff
@@ -208,27 +133,23 @@ def _begin_next_energy_step(self):
...
-        nominal energies The measured energies are then put into
-        relation to the nominal energies using
-        numpy.polynomial.polynomial.Polynomial. A polynomial of
-        first degree is most likely accurate enough for the
-        calibration, but the degree can be adjusted by changing
+        Measured energies are put into relation to the nominal
+        energies using numpy.polynomial.polynomial.Polynomial. A
+        polynomial of first degree is most likely accurate enough for
+        the calibration, but the degree can be adjusted by changing
         the integer value in the Polynomial.fit function.
```
michele-riva: Perhaps this comment about the order is not so clear in the docstring? I'm not sure there is anything that can be done at the user level to modify this order "externally"

src/viperleed/gui/measure/measurement/energy_calibration.py
```diff
@@ -208,27 +133,23 @@ def _begin_next_energy_step(self):
...
-        The measured energies are used as the x-coordinates
-        and the nominal energies are used as the y-coordinates.
-        The resulting polynomial is written into the config file
-        and used to calibrate the nominal energy via the
-        true_energy_to_setpoint() function to get the desired
-        output.
+        The measured energies are used as the x-coordinates and the
+        nominal energies are used as the y-coordinates. The resulting
+        polynomial is written into the config file and used to
+        calibrate the nominal energy via the true_energy_to_setpoint()
```
michele-riva: true_energy_to_setpoint is not a method of the measurement. Perhaps better be more specific here?

src/viperleed/gui/measure/widgets/fieldinfo.py
```diff
@@ -124,3 +124,36 @@ def _compose(self, label_text, tooltip):
         layout.addWidget(self._field_info)
         layout.setContentsMargins(0, 0, 0, 0)
         self.setLayout(layout)
+
+
+class InfoComboBox(qtw.QWidget):
+    """A QComboBox with an attached FieldInfo that displays item info text."""
```
michele-riva: You may want to describe a bit here how this works in a little more detail (needs items with user data from which the info_text attribute is pulled).

Side note: I think that, in principle, you could even just give tooltips to the items, without the need for an explicit FieldInfo. I think we do something like this in the ADC configuration of the viperino

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -1586,11 +1461,35 @@ def _on_image_saved(self, img_name):
...
+        ramp_name = self.settings.get('energies', 'ramp_type', fallback=None)
+        ramp_type = None
+        if ramp_name:
+            try:
+                ramp_type = base.class_from_name('classes', ramp_name)
+            except (RuntimeError, ValueError):
+                self.emit_error(QObjectSettingsErrors.INVALID_SETTINGS,
+                                'energies/ramp_type',
+                                f'Invalid string {ramp_name}.')
+                pass
```
michele-riva: pass not useful here

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,619 @@
...
+
+def _make_settings(**kwargs):
+    """Return a ViPErLEEDSettings."""
+    settings_dict = {'energies': kwargs} if kwargs else {}
+    settings = ViPErLEEDSettings()
+    settings.read_dict(settings_dict)
+    return settings
+
+
+#pylint: disable-next=too-few-public-methods
```
michele-riva:
```suggestion
# pylint: disable-next=too-few-public-methods
```

fix also others below

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,619 @@
...
+        assert linear_ramp._delta_energy == 0.0
+        assert linear_ramp._end_energy == 5.0
+
+
+class TestStepProfileFallbackAndGeneration:
+    """Tests for fallback logic and linear interpolation profile generation."""
+
+    def test_step_profile_falls_back_to_abrupt(self, linear_ramp):
+        """Check ABRUPT fallback branch when parsing as fractions fails."""
+        settings = _make_settings(step_profile='abrupt')
```
michele-riva: maybe you want to use a different string here, e.g., `'invalid'`?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,619 @@
...
+    def test_returns_linear_ramp_for_invalid_values(self, section):
+        """Check that invalid boolean values fall back to LinearEnergyRamp."""
+        settings = _make_settings(**section)
+        result = get_matching_energy_ramp(settings)
+        assert result is LinearEnergyRamp
+
+
+# pylint: disable=protected-access
+# pylint: disable=magic-value-comparison
+# pylint: disable=use-implicit-booleaness-not-comparison-to-zero
```
michele-riva: this seems to be used only in a few places. Perhaps better to disable-next where appropriate?

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,619 @@
...
+            (1.0, 2, 3.0, 2, 5.0, 2, 7.0, 2, 9.0, 2)
+            )
+
+    def test_get_linear_step_returns_empty_when_time_is_zero(self,
+                                                             linear_ramp):
+        """Check that 0 ms time intervals do not produce steps."""
+        linear_ramp.current_energy = 0.0
+        linear_ramp.current_energy = 10.0
+
+        assert linear_ramp._get_linear_step(20, 5) == ()
```
michele-riva, line 345: Add a comment that we do 5//20 == 0. Otherwise this test's docstring is confusing.

tests/gui/measure/classes/test_energyramp.py
```diff
@@ -0,0 +1,619 @@
...
+        """Check that non-sequence profiles are rejected as invalid."""
+        linear_ramp._set_step_profile(42)
+        assert linear_ramp._step_profile == (ABRUPT,)
+
+
+class TestRampSettingsWidgetsAndValidation:
+    """Tests for get_settings_widgets and ramp_settings_ok methods."""
+
+    # These are intentionally broken settings that we use to check that
+    # ramp_settings_ok correctly indentifies invalid configurations.
```
michele-riva: What I meant with my previous suggestion was to say what the items in the tuple are

michele-riva (changes requested):
Some more small things, but looks very good.

src/viperleed/gui/measure/classes/energyramp.py
FlorianDoerr: If you were to count it for labeling, yes. Since we do however not know how long this measurement will persist, we do the same as with the sawtooth profile, we return 0 when the number of steps cannot be determined upfront.

src/viperleed/gui/measure/measurement/energy_calibration.py
FlorianDoerr: We do not expose this, but potentially someone can just change deg=1 in Polynomial.fit to a higher degree.

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -198,6 +199,37 @@ def ramp_settings_ok(cls, energy_options):
         """
         return True, ''
 
+    @classmethod
+    def are_settings_invalid(cls, settings):
```
michele-riva: No use of `cls`. Perhaps a staticmethod is enough?

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -198,6 +199,37 @@ def ramp_settings_ok(cls, energy_options):
...
+            invalid setting.
+        """
+        ramp_name = settings.get('energies', 'ramp_type', fallback=None)
+        for ramp in ALL_ENERGY_RAMPS:
+            if ramp_name == ramp.__name__:
+                ramp_class = ramp
+                break
+        else:
+            ramp_class = get_ramp_from_settings(settings)
+        mandatory_settings = ramp_class._mandatory_settings
```
michele-riva, line 227: pylint will complain about accessing a private attribute here. I think it can be safely disabled commenting that ramp_class is always a subclass of EnergyRampABC

src/viperleed/gui/measure/classes/energyramp.py
```diff
@@ -608,3 +655,7 @@ def increment_energy(self):
     def ramp_finished(self):
         """Return whether the energy ramp has been finished."""
         return False  # An endless energy ramp is never finished.
+
+
+ALL_ENERGY_RAMPS = (LinearEnergyRamp, SawtoothEnergyRamp,
+                    ConstantEnergyRamp)
```
michele-riva: perhaps better to have one per line and alphabetically sorted, like so:

```suggestion
ALL_ENERGY_RAMPS = (
    ConstantEnergyRamp,
    LinearEnergyRamp,
    SawtoothEnergyRamp,
    )
```

Makes it easier to add more at later stages

michele-riva (changes requested):
Looks very good, only a couple very minor points left

michele-riva (approved):
Looks good

PR #523, opened 2026-03-19
fix: load SUPERLATTICE to main params

fkraushofer:
The SUPERLATTICE from domains needs to be copied to the main rparams object before generating the BEAMLIST; otherwise, the BEAMLIST will assume (1x1) and all fractional beams will be discarded.

Domain calculations are currently completely non-functional without this fix. This is confirmed to work [tested on hematite (012)-(2x1)].


michele-riva (approved):
Makes perfect sense. Interesting that this did not pop up earlier (the old `runBeamgen` also used `rp.SUPERLATTICE`).

PR #500, opened 2025-11-19
Python rfactor

alexmiame:
This PR adds a native Python R factor calculation to viperleed. An effort is made to keep the code agnostic to the numerical library, such that the same routines can be used with NumPy and in JAX (used in the viperleed-jax plugin, see also #333).

PR #526, opened 2026-03-31
Robust r factor selection

fkraushofer:
This branch was originally merged into https://github.com/viperleed/viperleed/tree/viperleed-jax-plugin only, but is also required in the upstream branches, in particular for #500.


fkraushofer:
Actually, upon review, it seems almost impossible to merge #500 without also merging #333. I'm closing this one again, we'll have to see how to manage #333.

EDIT: Actually, #500 is fine. It is this one which is problematic. I'm trying to do a manual cherry-pick to fix.

PR #536, opened 2026-04-02
doc,fix: update copyright to 2026

alexmiame:
Updates the copyright notices to 2026 in calc.


michele-riva (changes requested):
@alexmiame: legally, copyright dates should only be edited on those files that have been edited during the corresponding year. I see that some of the files that this PR modifies have not been edited in 2026. Can you revert the changes for these files?

PR #543, opened 2026-04-23
Docs in pyproject

fkraushofer:
Adds `[docs]` as an optional install to the viperleed `pyproject.toml`. The alternative would be to make a docs/requirements.txt, but I feel that's a bit less clean. Either way, this allows developers to quickly and easily set up the environment such that they can immediately build the docs.


michele-riva:
There is already a requirements/doc.txt

fkraushofer:
Ah, oops, didn't see that. Never mind then!

PR #544, opened 2026-04-23
Rename R_FACTOR_SMOOTH

### LINKED ISSUE #493: Rename R_FACTOR_SMOOTH to R_FACTOR_SMOOTHING
Since the [new R-factor](https://arxiv.org/abs/2511.05448v1) proposed by @schmid-iap is called the "smooth" R-factor R_S, I find having another parameter called [R_FACTOR_SMOOTH ](https://www.viperleed.org/stable/content/calc/parameters/rfactorsmooth.html#rfactorsmooth) (which controls the 3-point-smoothing as implemented in TensErLEED) slightly confusing. 

I propose to rename it to R_FACTOR_SMOOTHING, while keeping R_FACTOR_SMOOTH as an alias for backwards compatibility. The choice of R-factor (to switch between Pendry and others) is a different parameter, called [R_FACTOR_TYPE](https://www.viperleed.org/stable/content/calc/parameters/rfactortype.html#rfactortype).

fkraushofer:
Closes #493


doc/content/calc/parameters/rfactorsmooth.rst
```diff
@@ -2,13 +2,17 @@
 
 .. _rfactorsmooth:
 
-R_FACTOR_SMOOTH
+R_FACTOR_SMOOTHING
 ===============
```
michele-riva:
```suggestion
==================
```

michele-riva: prevents sphinx warnings

doc/content/calc/parameters/rfactorsmooth.rst
```diff
@@ -2,13 +2,17 @@
...
 
-R_FACTOR_SMOOTH defines the number of times that experimental |IV| curves
+R_FACTOR_SMOOTHING defines the number of times that experimental |IV| curves
 will be smoothed by 3-point smoothing during |R-factor| calculations, both
```
michele-riva, line 9: Perhaps worth mentioning that this only applies to TensErLEED?

michele-riva (changes requested):
tiny suggestions, but looks good otherwise

michele-riva (approved):
Looks good!

PR #549, opened 2026-05-14
Report quantity mismatch

FlorianDoerr:
Catch ValueError when quantity mismatch happens and report the error to the user.

### DIFF
--- a/src/viperleed/gui/measure/widgets/collapsibleviews.py
+++ b/src/viperleed/gui/measure/widgets/collapsibleviews.py
@@ -544,4 +544,4 @@ class CollapsibleControllerView(CollapsibleDeviceView):
             self._quantity_selector.set_quantities(quantities)
         except ValueError as err:
             emit_error(self, QObjectSettingsErrors.INVALID_SETTINGS,
-                        'quantities', str(err))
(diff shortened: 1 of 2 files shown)


src/viperleed/gui/measure/widgets/collapsibleviews.py
```diff
@@ -474,7 +484,11 @@ def _build_device_settings_widgets(self):
         layout = self._frame.layout().itemAt(2).widget().layout()
         self._quantity_selector = QuantitySelector(settings)
         if self._quantities_to_set:
-            self._quantity_selector.set_quantities(self._quantities_to_set)
+            try:
+                self._quantity_selector.set_quantities(self._quantities_to_set)
+            except ValueError as err:
+                emit_error(self, QObjectSettingsErrors.INVALID_SETTINGS,
+                           'quantities', str(err))
```
michele-riva: This new bit of code is a bit a repetition of the one in set_quantities. Consider whether you could just call set_quantities here.

michele-riva (changes requested):
Looks good, only one small suggestion

michele-riva (approved):
Looks good!

PR #553, opened 2026-07-05
Test the workflow of the Opencode bot

FlorianDoerr:
This PR is here to test the Opencode bot.

### DIFF
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -16,7 +16,7 @@ jobs:
       id-token: write
       pull-requests: write
       issues: write
-      contents: read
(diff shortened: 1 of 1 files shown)


FlorianDoerr:
@michele-riva I am afraid the workflow is not discovered if it is not present in the master branch as outlined in this comment: https://github.com/orgs/community/discussions/180806#discussioncomment-15120439
I assume actions are enabled?

michele-riva:
I've updated some permissions, as per opencode request. Perhaps that has unlocked it?

FlorianDoerr:
> I've updated some permissions, as per opencode request. Perhaps that has unlocked it?

Doesn't look like it.

PR #555, opened 2026-07-21
Fix mandatory and missing settings check

FlorianDoerr:
Fix `misses_settings` and `_check_mandatory_settings` of `ViPErLEEDSettings`. The previous implementation would falsely report missing section/option pairs as just missing sections if the section was missing. Furthermore, the section/option/value handling was broken.

### DIFF
--- a/src/viperleed/gui/measure/classes/settings.py
+++ b/src/viperleed/gui/measure/classes/settings.py
@@ -442,28 +442,32 @@ class ViPErLEEDSettings(AliasConfigParser):
                 raise TypeError(f'Invalid mandatory setting {setting}. with '
                                 f'length {length}. Expected length <= 3.')
 
+            # First, check if settings are broken without reporting yet.
             settings_broken = False
-            if not self.has_section(setting[0]):
+            section, option, admissible = tuple(setting) + (None,) * (3-length)
+            if not self.has_section(section):
                 settings_broken = True
-            elif length > 1 and not self.has_option(setting[0], setting[1]):
+            elif length > 1 and not self.has_option(section, option):
                 settings_broken = True
-            elif length > 2 and self[setting[0]][setting[1]] not in setting[2]:
+            elif length > 2 and self[section][option] not in admissible:
                 settings_broken = True
 
             if not settings_broken:
                 continue
 
+            # Then report broken settings. This two-step approach ensures
+            # the entire setting is reported regardless of which part failed.
             # (<section>,)
             if length == 1:
-                invalid_settings.append(setting[0])
+                invalid_settings.append(section)
                 continue
             # (<section>, <option>)
             if length == 2:
-                invalid_settings.append(f'{setting[0]}/{setting[1]}')
+                invalid_settings.append(f'{section}/{option}')
                 continue
             # (<section>, <option>, <admissible>)
             invalid_settings.append('/'.join(setting[:2]) + ' not one of ' +
-                                    ', '.join(setting[2]))
+                                    ', '.join(admissible))
         return invalid_settings
 
     def read(self, filenames, encoding=None):
--- a/src/viperleed/gui/measure/classes/settings.py
+++ b/src/viperleed/gui/measure/classes/settings.py
@@ -873,7 +873,7 @@ class SystemSettings(ViPErLEEDSettings):
         # We're missing settings. Let's add them back...
         for missing in invalid:
             # 1. Strip off the value-error suffix if present
-            clean_keys = missing.split(' not one of ')[0]
+            clean_keys, *_ = missing.split(' not one of ')
             if '/' in clean_keys:
                 section, option = clean_keys.split('/')
             else:
--- a/src/viperleed/gui/measure/classes/settings.py
+++ b/src/viperleed/gui/measure/classes/settings.py
@@ -437,32 +437,33 @@ class ViPErLEEDSettings(AliasConfigParser):
         """
         invalid_settings = []
         for setting in required_settings:
-            if not setting or len(setting) > 3:
-                raise TypeError(f'Invalid mandatory setting {setting}. '
-                                f'with length {len(setting)}. Expected '
-                                'length <= 3.')
+            length = len(setting)
+            if not setting or length > 3:
+                raise TypeError(f'Invalid mandatory setting {setting}. with '
+                                f'length {length}. Expected length <= 3.')
+
+            settings_broken = False
+            if not self.has_section(setting[0]):
+                settings_broken = True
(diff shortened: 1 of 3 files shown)


tests/gui/measure/classes/test_settings.py
```diff
@@ -291,6 +294,96 @@ def test_hidden_folder_and_settings_creation(self, tmp_path, mocker):
...
+        mocker.patch.object(sys_settings, 'update_file')
+        return sys_settings
+
+    # pylint: disable=protected-access
+    def test_auto_fills_missing_non_null_settings(self, settings):
+        """Check that missing settings are auto-created."""
+        # Define requirements
+        settings._SystemSettings__non_null = [('SecA', 'opt1')]
+        settings._SystemSettings__non_mandatory = [('SecB',)]
+        settings._SystemSettings__mandatory = []
```
michele-riva: Use mocker.patch here to prevent permanent edits to class attributes: they are restored as soon as the test finishes. Then the pylint disables can also be moved locally to the call to `_check_mandatory..`

tests/gui/measure/classes/test_settings.py
```diff
@@ -291,6 +294,96 @@ def test_hidden_folder_and_settings_creation(self, tmp_path, mocker):
         # Patch QSettings.allKeys to force creation of settings folder.
         mocker.patch('PyQt5.QtCore.QSettings.allKeys', return_value=None)
         sys_settings = SystemSettings()
+        # pylint: disable-next=protected-access
         settings_path = Path(sys_settings._sys_qsettings.fileName()).resolve()
         assert settings_path.parent.is_dir()
         assert settings_path.is_file()
+
+
+class TestCheckMandatorySettings:
```
michele-riva: Is there a reason why these tests are not within the TestSystemSettings class?

tests/gui/measure/classes/test_settings.py
```diff
@@ -291,6 +294,96 @@ def test_hidden_folder_and_settings_creation(self, tmp_path, mocker):
...
+        settings._SystemSettings__non_null = []
+        settings._SystemSettings__non_mandatory = []
+        settings._SystemSettings__mandatory = [('MandatorySec', 'opt')]
+
+        with pytest.raises(RuntimeError, match='MandatorySec/opt'):
+            settings._check_mandatory_settings()
+    # pylint: enable=protected-access
+
+
+class TestMissesSettings:
```
michele-riva: These tests would fit better under a TestViPErLEEDSettings class (unless they would also apply to objects higher up in the class hierarchy. Then I would move them there)

tests/gui/measure/classes/test_settings.py
```diff
@@ -291,6 +294,96 @@ def test_hidden_folder_and_settings_creation(self, tmp_path, mocker):
...
+        """Check that no invalid settings are returned when all exist."""
+        parser = ViPErLEEDSettings(cls_name='Foo')
+        parser.read_dict({'Sec': {'opt1': 'val1', 'opt2': 'a'}})
+
+        invalid = parser.misses_settings(
+            ('Sec',),
+            ('Sec', 'opt1'),
+            ('Sec', 'opt2', ['a', 'b'])
+        )
+        assert invalid == []    # pylint: disable=C1803
```
michele-riva:
```suggestion
        assert not invalid
```

I think for the purpose of this test to succeed, it doesn't matter that the return value is an empty list, but only that there is nothing invalid

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -437,32 +437,33 @@ def misses_settings(self, *required_settings):
...
-                raise TypeError(f'Invalid mandatory setting {setting}. '
-                                f'with length {len(setting)}. Expected '
-                                'length <= 3.')
+            length = len(setting)
+            if not setting or length > 3:
+                raise TypeError(f'Invalid mandatory setting {setting}. with '
+                                f'length {length}. Expected length <= 3.')
 
-            # (<section>,)
+            settings_broken = False
```
michele-riva, line 446: Perhaps it would be more readable to "unpack" `setting` to a `section, option, admissible` triplet (with missing parts set to `None`) rather than using setting[0], [1], [2]?

```suggestion
            section, option, admissible = tuple(setting) + (None,) * (3-length)
            settings_broken = False
```

michele-riva: I would also advise leaving a comment as to why this is done in two passes, as the commit message was not clear

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -871,10 +872,18 @@ def _check_mandatory_settings(self):
 
         # We're missing settings. Let's add them back...
         for missing in invalid:
-            section, option = missing.split('/')
+            # 1. Strip off the value-error suffix if present
+            clean_keys = missing.split(' not one of ')[0]
```
michele-riva:
```suggestion
            clean_keys, *_ = missing.split(' not one of ')
```

michele-riva (changes requested):
Looks good, some small suggestions left, though

tests/gui/measure/classes/test_settings.py
```diff
@@ -291,6 +328,61 @@ def test_hidden_folder_and_settings_creation(self, tmp_path, mocker):
...
+            # pylint: disable-next=protected-access
+            settings._check_mandatory_settings()
+
+
+class TestViPErLEEDSettings:
+    """Tests for the ViPErLEEDSettings class."""
+
+    def test_all_settings_valid(self):
+        """Check that no invalid settings are returned when all exist."""
+        parser = ViPErLEEDSettings(cls_name='Foo')
```
michele-riva: Do we really need to pass a cls_name to all of these? I don't think we're using the aliases in these tests

tests/gui/measure/classes/test_settings.py
```diff
@@ -291,6 +328,61 @@ def test_hidden_folder_and_settings_creation(self, tmp_path, mocker):
...
+        parser.read_dict({'Sec': {'opt1': 'val1', 'opt2': 'a'}})
+
+        invalid = parser.misses_settings(
+            ('Sec',),
+            ('Sec', 'opt1'),
+            ('Sec', 'opt2', ['a', 'b'])
+        )
+        assert not invalid
+
+    def test_missing_section(self):
```
michele-riva: sort alphabetically

michele-riva (changes requested):
Looks good

michele-riva (approved):

PR #556, opened 2026-07-23
Catch temperature conversion error

FlorianDoerr:
Catch the ValueError and emit error message rather than crashing.

### DIFF
--- a/src/viperleed/gui/measure/controller/viperinocontroller.py
+++ b/src/viperleed/gui/measure/controller/viperinocontroller.py
@@ -1165,9 +1165,10 @@ class ViPErinoController(MeasureControllerABC):
                 ]
         except ValueError as err:
             self.measurements[QuantityInfo.TEMPERATURE] = tc_voltages
-            self.emit_error(ViPErinoErrors.CANNOT_CONVERT_THERMOCOUPLE,
-                            f'\nInfo: {err}.')
-
+            self.emit_error(
+                ViPErinoErrors.CANNOT_CONVERT_THERMOCOUPLE, f'\nInfo: {err}. '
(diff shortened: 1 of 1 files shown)


src/viperleed/gui/measure/controller/viperinocontroller.py
```diff
@@ -1158,10 +1158,16 @@ def __convert_thermocouple_voltages(self):
...
-            ]
+        try:
+            self.measurements[QuantityInfo.TEMPERATURE] = [
+                self.thermocouple.temperature(v, t0)
+                for v, t0 in zip(tc_voltages, cjc_temperatures)
+                ]
+        except ValueError as err:
+            self.measurements[QuantityInfo.TEMPERATURE] = tc_voltages
+            self.emit_error(ViPErinoErrors.CANNOT_CONVERT_THERMOCOUPLE,
+                            f'\nInfo: {err}.')
```
michele-riva:
```suggestion
        except ValueError as exc:
            self.measurements[QuantityInfo.TEMPERATURE] = tc_voltages
            self.emit_error(ViPErinoErrors.CANNOT_CONVERT_THERMOCOUPLE,
                            f'\nInfo: {exc}.')
```

I think we always used `exc` for exceptions

michele-riva: I would also give users some more suggestions:
1. they may have the thermocouple wires connected incorrectly
2. if they're correct, then they should open an issue

michele-riva (changes requested):
Some small suggestions on the error message

michele-riva (approved):
Looks good

PR #557, opened 2026-07-24
Correct minimum width for device lists

FlorianDoerr:
Old:
image
New:
image

### DIFF
--- a/src/viperleed/gui/measure/widgets/collapsiblelists.py
+++ b/src/viperleed/gui/measure/widgets/collapsiblelists.py
@@ -271,7 +271,7 @@ class CollapsibleDeviceList(CollapsibleList):
         widget = self.widget()
         width = widget.minimumSizeHint().width()
         view_widths = (
-            view.minimumSizeHint().width() for view in self.views.keys()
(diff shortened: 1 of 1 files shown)


src/viperleed/gui/measure/widgets/collapsiblelists.py
```diff
@@ -259,6 +264,20 @@ def _make_top_items(self):
...
         self._layout.insertLayout(1, top_labels)
 
+    def _required_width(self):
+        """Return minimum pixel width required to display all contents."""
+        if not self.widget():
+            return 10
+        widget = self.widget()
+        width = widget.minimumSizeHint().width()
+        view_widths = (
+            view.minimumSizeHint().width() for view in self.views.keys()
```
michele-riva:
```suggestion
            view.minimumSizeHint().width() for view in self.views
```

It doesn't matter that it's a dict. We only need it to be iterable for this

michele-riva (changes requested):
Looks good!

michele-riva (approved):

PR #559, opened 2026-08-11
Set energy from gui

FlorianDoerr:
Allow setting the beam energy from the measurement GUI and add a way to select the primary controller for this operation.

### DIFF
--- a/src/viperleed/gui/measure/_defaults/_system_settings.ini
+++ b/src/viperleed/gui/measure/_defaults/_system_settings.ini
@@ -7,3 +7,4 @@ firmware =
 
 [DEVICES]
 live_detection =
+primary_controller =
--- a/src/viperleed/gui/measure/_defaults/_system_settings.ini
+++ b/src/viperleed/gui/measure/_defaults/_system_settings.ini
@@ -7,4 +7,4 @@ firmware =
 
 [DEVICES]
 live_detection =
-primary_controller =
+controller =
--- a/src/viperleed/gui/measure/_defaults/_system_settings.ini
+++ b/src/viperleed/gui/measure/_defaults/_system_settings.ini
@@ -7,4 +7,4 @@ firmware =
 
 [DEVICES]
 live_detection =
-controller =
+default_controller =
--- a/src/viperleed/gui/measure/controller/abc.py
+++ b/src/viperleed/gui/measure/controller/abc.py
@@ -1193,7 +1193,7 @@ class MeasureControllerABC(ControllerABC):
         tip = ("<nobr>The number of measurements the controller should"
                "</nobr> average over before returning a value to the PC.")
         handler.add_option('measurement_settings', 'nr_samples',
-                           handler_widget=widget, display_name='No. samples',
+                           handler_widget=widget, display_name='Nr. samples',
                            tooltip=tip)
         return handler
 
--- a/src/viperleed/gui/measure/controller/abc.py
+++ b/src/viperleed/gui/measure/controller/abc.py
@@ -695,7 +695,7 @@ class ControllerABC(DeviceABC):
              '</nobr> to reach a stable value after a new energy has '
              'been set. This should be calibrated for a typical step '
              'size (e.g., 0.5 eV).'),
-            ('first_settle_time', '1st energy settle time',
+            ('first_settle_time', 'E start settle time',
              '<nobr>The time interval required for the true beam '
              'energy</nobr> to reach a stable value when the first energy '
              'of a ramp is set. This is usually significantly longer than '
--- a/src/viperleed/gui/measure/controller/abc.py
+++ b/src/viperleed/gui/measure/controller/abc.py
@@ -27,6 +27,7 @@ from viperleed.gui.measure.classes.abc import DeviceABCErrors
 from viperleed.gui.measure.classes.abc import NO_HARDWARE_INTERFACE
 from viperleed.gui.measure.classes.abc import QObjectSettingsErrors
 from viperleed.gui.measure.classes.datapoints import QuantityInfo
+from viperleed.gui.measure.classes.energyramp import START_E_NAME
 from viperleed.gui.measure.classes.settings import NotASequenceError
 from viperleed.gui.measure.dialogs.settingsdialog import SettingsTag
 from viperleed.gui.measure.hardwarebase import disconnected_signal
@@ -695,7 +696,7 @@ class ControllerABC(DeviceABC):
              '</nobr> to reach a stable value after a new energy has '
              'been set. This should be calibrated for a typical step '
              'size (e.g., 0.5 eV).'),
-            ('first_settle_time', 'E start settle time',
+            ('first_settle_time', START_E_NAME + ' settle time',
              '<nobr>The time interval required for the true beam '
              'energy</nobr> to reach a stable value when the first energy '
              'of a ramp is set. This is usually significantly longer than '
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -165,7 +165,7 @@ class EnergySetter(qtw.QWidget):
 
         Returns
         -------
-        ControllerABC or None
+        primary_ctrl : ControllerABC or None
             The controller instance, or None if creation failed.
 
         Emits
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -106,8 +106,9 @@ class EnergySetter(qtw.QWidget):
         if not primary_path.is_file():
             qtw.QMessageBox.warning(
                 self, 'Controller File Missing',
-                f'The primary controller settings file no longer exists:\n{primary_path}\n'
-                'Please select a new primary controller from the "Devices" menu.'
+                'The primary controller settings file no longer exists:'
+                f'\n{primary_path}\n Please select a new primary '
+                'controller from the "Devices" menu.'
             )
             self.set_energy.setChecked(False)
             return
@@ -198,7 +199,7 @@ class EnergySetter(qtw.QWidget):
             self.set_energy.setChecked(False)
             return
 
-        # Create event loop to wait for completion
+        # Create event loop to wait for completion.
         loop = qtc.QEventLoop()
 
         def on_finished():
@@ -213,10 +214,10 @@ class EnergySetter(qtw.QWidget):
                 f'Failed to set energy:\n{error_info}'
             )
 
-        # Connect to serial's connection_changed to know when done
+        # Connect to serial's connection_changed to know when done.
         primary_ctrl.error_occurred.connect(on_error)
 
-        # Set up timeout in case no response arrives
+        # Set up timeout in case no response arrives.
         timeout_timer = qtc.QTimer()
         timeout_timer.setSingleShot(True)
         timeout_timer.setInterval(5000)  # 5 second timeout
@@ -229,7 +230,7 @@ class EnergySetter(qtw.QWidget):
             loop.quit()
         ))
 
-        # Connect to busy_changed - when it goes False, operation is complete
+        # Connect to busy_changed - when it goes False, operation is complete.
         primary_ctrl.serial.busy_changed.connect(
             lambda busy: (
                 timeout_timer.stop(),
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -256,8 +256,13 @@ class EnergySetter(qtw.QWidget):
         self._operation_in_progress = False
         self._timeout_timer.stop()
 
+        # If set energy is no longer toggled, we want to disconnect the ctrl.
+        if not self.set_energy.isChecked():
+            self.cleanup_primary_controller()
+            return
+
         # If a new energy value was queued during the operation, process it.
-        if self._pending_energy is not None and self.set_energy.isChecked():
+        if self._pending_energy is not None:
             energy = self._pending_energy
             self._pending_energy = None
             self._set_energy(energy)
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -239,7 +239,7 @@ class EnergySetter(qtw.QWidget):
         if self._primary_controller is None:
             return
         base.safe_disconnect(self._primary_controller.error_occurred,
-                             self._on_error,)
+                             self._on_error)
         base.safe_disconnect(self._primary_controller.serial.busy_changed,
                              self._on_ctrl_finished)
         self._primary_controller.disconnect_()
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -38,19 +38,17 @@ class EnergySetterErrors(base.ViPErLEEDErrorEnum):
     CONTROLLER_CONNECTION_FAILED = (2003, 'Could not connect to the '
                                     'controller. Please check that the '
                                     'device is available.')
-    SET_ENERGY_FAILED = (2004, 'Failed to set energy:\n{}')
-    SET_ENERGY_TIMEOUT = (2005, 'No response from controller. Check '
+    SET_ENERGY_TIMEOUT = (2004, 'No response from controller. Check '
                           'connection.')
 
 
 class EnergySetter(qtw.QWidget):
-    """Widget for setting LEED energy without data acquisition.
+    """Widget for setting LEED energies without data acquisition.
 
     This widget provides a checkbox and energy input field that allow
-    users to set the beam energy on the primary controller without
-    starting a measurement.
-
+    users to set the beam energy.
     """
+
     error_occurred = qtc.pyqtSignal(tuple)
 
     def __init__(self, **kwargs):
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -112,7 +112,20 @@ class EnergySetter(qtw.QWidget):
 
     @qtc.pyqtSlot(int)
     def _on_set_energy_toggled(self, state):
-        """Handle checkbox state change."""
+        """Handle checkbox state change.
+
+        Parameters
+        ----------
+        state : qtc.Qt.CheckState
+            The checked state of the QCheckBox.
+
+        Emits
+        -----
+        EnergySetterErrors.NO_PRIMARY_CONTROLLER
+            If no primary path was given.
+        EnergySetterErrors.CONTROLLER_FILE_MISSING
+            If the primary path does not point to a file.
+        """
         self.energy_input.setEnabled(state == qtc.Qt.Checked)
         if state != qtc.Qt.Checked:
             # Checkbox unchecked, set energy to zero before cleaning up.
@@ -175,7 +188,7 @@ class EnergySetter(qtw.QWidget):
 
         # Create new controller instance.
         try:
-            primary_ctrl = self._make_primary()
+            primary_ctrl = self._make_primary_controller()
         except (NoSettingsError, ValueError) as err:
             base.emit_error(self, EnergySetterErrors.CONTROLLER_LOAD_FAILED,
                             err)
@@ -195,7 +208,7 @@ class EnergySetter(qtw.QWidget):
 
         return primary_ctrl
 
-    def _make_primary(self):
+    def _make_primary_controller(self):
         """Make and return a new primary controller.
 
         Returns
@@ -234,13 +247,29 @@ class EnergySetter(qtw.QWidget):
 
     @qtc.pyqtSlot(tuple)
     def _on_error(self, error_info):
-        """Handle controller errors."""
+        """Handle controller errors.
+
+        Parameters
+        ----------
+        error_info : tuple
+
+        Emits
+        -----
+        error_occurred
+            With error_info on the error that occurred.
+        """
         self.error_occurred.emit(error_info)
         self._flush()
 
     @qtc.pyqtSlot()
     def _on_timeout(self):
-        """Handle timeout."""
+        """Handle timeout.
+
+        Emits
+        -----
+        EnergySetterErrors.SET_ENERGY_TIMEOUT
+            When a timeout happened.
+        """
         base.emit_error(self, EnergySetterErrors.SET_ENERGY_TIMEOUT)
         self._flush()
 
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -60,7 +60,7 @@ class EnergySetter(qtw.QWidget):
         self.set_energy = qtw.QCheckBox('Set energy')
         self._primary_path = None
         self._primary_controller = None
-        self._pending_energy = None     # Store energy value during operation
+        self._pending_energy = None
         self._operation_in_progress = False
 
         self._timeout_timer = qtc.QTimer(self)
@@ -82,7 +82,7 @@ class EnergySetter(qtw.QWidget):
 
     @primary_path.setter
     def primary_path(self, primary):
-        """Set the primary path to the primary controller."""
+        """Set the path to the primary controller."""
         self._primary_path = Path(primary)
 
     def _compose(self):
@@ -268,7 +268,7 @@ class EnergySetter(qtw.QWidget):
             self._set_energy(energy)
 
     def set_enabled(self, enable):
-        """Switch enabled status of buttons."""
+        """Switch enabled status of widgets."""
         enable &= bool(self.primary_path)
         self.set_energy.setEnabled(enable)
         self.energy_input.setEnabled(enable and self.set_energy.isChecked())
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -81,7 +81,10 @@ class EnergySetter(qtw.QWidget):
     @primary_path.setter
     def primary_path(self, primary):
         """Set the path to the primary controller."""
-        self._primary_path = Path(primary)
+        if primary:
+            self._primary_path = Path(primary)
+        else:
+            self._primary_path = None
 
     def _compose(self):
         """Set up the user interface."""
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -153,6 +153,13 @@ class EnergySetter(qtw.QWidget):
         -------
         ControllerABC or None
             The controller instance, or None if creation failed.
+
+        Emits
+        -----
+        EnergySetterErrors.CONTROLLER_LOAD_FAILED
+            If making the primary controller failed.
+        EnergySetterErrors.CONTROLLER_CONNECTION_FAILED
+            If connecting the primary controller failed.
         """
         # If we already have a controller for this path, reuse it.
         if self._primary_controller is not None:
@@ -163,21 +170,13 @@ class EnergySetter(qtw.QWidget):
                 if not self._primary_controller.connected:
                     self._primary_controller.connect_()
                 return self._primary_controller
-            else:
-                # Different controller, clean up the old one.
-                self.cleanup_primary_controller()
+            # Different controller, clean up the old one.
+            self.cleanup_primary_controller()
 
         # Create new controller instance.
         try:
-            ctrl_settings = ViPErLEEDSettings()
-            ctrl_settings.read(self.primary_path)
-            ctrl_cls_name = ctrl_settings.get('controller', 'controller_class')
-            ctrl_settings.prepare_aliases(ctrl_cls_name)
-            ctrl_cls = base.class_from_name('controller', ctrl_cls_name)
-            address = ctrl_settings.get('controller', 'address')
-            primary_ctrl = ctrl_cls(settings=ctrl_settings, address=address,
-                                    sets_energy=True)
-        except (NoSettingsError, ValueError, KeyError) as err:
+            primary_ctrl = self._make_primary()
+        except (NoSettingsError, ValueError) as err:
             base.emit_error(self, EnergySetterErrors.CONTROLLER_LOAD_FAILED,
                             err)
             return None
@@ -194,6 +193,31 @@ class EnergySetter(qtw.QWidget):
             self.cleanup_primary_controller()
             return None
 
+        return primary_ctrl
+
+    def _make_primary(self):
+        """Make and return a new primary controller.
+
+        Returns
+        -------
+        primary_ctrl : ControllerABC
+            A controller capable of setting the energy.
+
+        Raises
+        ------
+        NoSettingsError
+            If the settings file cannot be read.
+        ValueError
+            If ctrl_cls_name was not found.
+        """
+        ctrl_settings = ViPErLEEDSettings()
+        ctrl_settings.read(self.primary_path)
+        ctrl_cls_name = ctrl_settings.get('controller', 'controller_class')
+        ctrl_settings.prepare_aliases(ctrl_cls_name)
+        ctrl_cls = base.class_from_name('controller', ctrl_cls_name)
+        address = ctrl_settings.get('controller', 'address')
+        primary_ctrl = ctrl_cls(settings=ctrl_settings, address=address,
+                                sets_energy=True)
         return primary_ctrl
 
     def cleanup_primary_controller(self):
--- a/src/viperleed/gui/measure/energysetter.py
+++ b/src/viperleed/gui/measure/energysetter.py
@@ -1,0 +1,249 @@
+"""Module energysetter of viperleed.gui.measure.
+
+Defines the EnergySetter widget for setting LEED energy without
+starting a measurement.
+"""
+
+__authors__ = (
+    'Michele Riva (@michele-riva)',
+    'Florian Dörr (@FlorianDoerr)',
+    )
+__copyright__ = 'Copyright (c) 2019-2026 ViPErLEED developers'
+__created__ = '2026-07-31'
+__license__ = 'GPLv3+'
+
+from pathlib import Path
+
+from PyQt5 import QtCore as qtc
+from PyQt5 import QtWidgets as qtw
+
+from viperleed.gui.measure import hardwarebase as base
+from viperleed.gui.measure.classes.settings import NoSettingsError
+from viperleed.gui.measure.classes.settings import ViPErLEEDSettings
+from viperleed.gui.measure.widgets.spinboxes import SteppingDoubleSpinBox
+from viperleed.gui.widgets.lib import AllGUIFonts
+
+
+class EnergySetter(qtw.QWidget):
+    """Widget for setting LEED energy without data acquisition.
+
+    This widget provides a checkbox and energy input field that allow
+    users to set the beam energy on the primary controller without
+    starting a measurement.
+
+    """
+    error_occurred = qtc.pyqtSignal(tuple)
+
+    def __init__(self, **kwargs):
+        """Initialize the EnergySetter widget."""
+        super().__init__(**kwargs)
+        self.energy_input = SteppingDoubleSpinBox()
+        self.set_energy = qtw.QCheckBox('Set energy')
+        self.primary_path = ''
+        self._primary_controller = None
+
+        self._compose()
+        self._connect()
+
+    @property
+    def setting_energy(self):
+        """Return whether the energy setter is setting energies or not."""
+        return self.set_energy.checkState() == qtc.Qt.Checked
+
+    def _compose(self):
+        """Set up the user interface."""
+        layout = qtw.QGridLayout(self)
+        layout.setContentsMargins(0, 0, 0, 0)
+        for widget in (self.set_energy, self.energy_input):
+            widget.setFont(AllGUIFonts().buttonFont)
+            widget.ensurePolished()
+            widget.setEnabled(False)
+        layout.addWidget(self.set_energy, 0, 0)
+        self.energy_input.setDecimals(1)
+        self.energy_input.setRange(0.0, 1000.0)
+        self.energy_input.setSingleStep(0.5)
+        self.energy_input.setValue(0.0)
+        self.energy_input.setSuffix(' eV')
+        layout.addWidget(self.energy_input, 0, 1)
+
+        self.setLayout(layout)
+
+    def _connect(self):
+        """Connect internal signals and slots."""
+        self.set_energy.stateChanged.connect(self._on_set_energy_toggled)
+        self.energy_input.editingFinished.connect(self._on_energy_changed)
+        self.energy_input.valueChanged.connect(self._on_energy_changed)
+
+    @qtc.pyqtSlot(int)
+    def _on_set_energy_toggled(self, state):
+        """Handle checkbox state change."""
+        if state != qtc.Qt.Checked:
+            # Checkbox unchecked, set energy to zero before cleaning up.
+            if self.primary_path:
+                primary_path = Path(self.primary_path)
+                if primary_path.is_file():
+                    self._set_energy(primary_path, 0.0)
+        self.cleanup_primary_controller()
+        # When checked, controller will be created on first energy change.
+
+    @qtc.pyqtSlot()
+    @qtc.pyqtSlot(float)
+    def _on_energy_changed(self, *_):
+        """Handle energy value change."""
+        if self.set_energy.checkState() != qtc.Qt.Checked:
+            return
+
+        if not self.primary_path:
+            qtw.QMessageBox.warning(
+                self, 'No Controller Available',
+                'No primary controller configured. Please select one from '
+                'the "Devices" menu using "Select Primary Controller...".'
+            )
+            self.set_energy.setChecked(False)
+            return
+
+        primary_path = Path(self.primary_path)
+        if not primary_path.is_file():
+            qtw.QMessageBox.warning(
+                self, 'Controller File Missing',
+                f'The primary controller settings file no longer exists:\n{primary_path}\n'
+                'Please select a new primary controller from the "Devices" menu.'
+            )
+            self.set_energy.setChecked(False)
+            return
+
+        energy = self.energy_input.value()
+        self._set_energy(primary_path, energy)
+
+    def _get_primary_controller(self, primary_path):
+        """Get or create persistent primary controller instance.
+
+        Parameters
+        ----------
+        primary_path : Path
+            Path to the controller settings file.
+
+        Returns
+        -------
+        ControllerABC or None
+            The controller instance, or None if creation failed.
+        """
+        # If we already have a controller for this path, reuse it.
+        if self._primary_controller is not None:
+            ctrl_settings = self._primary_controller.settings
+            if (ctrl_settings.last_file and
+                ctrl_settings.last_file == primary_path):
+                # Same controller, ensure it's connected.
+                if not self._primary_controller.connected:
+                    self._primary_controller.connect_()
+                return self._primary_controller
+            else:
+                # Different controller, clean up the old one.
+                self.cleanup_primary_controller()
+
+        # Create new controller instance.
+        try:
+            ctrl_settings = ViPErLEEDSettings()
+            ctrl_settings.read(primary_path)
+            ctrl_cls_name = ctrl_settings.get('controller', 'controller_class')
+            ctrl_settings.prepare_aliases(ctrl_cls_name)
+            ctrl_cls = base.class_from_name('controller', ctrl_cls_name)
+            address = ctrl_settings.get('controller', 'address')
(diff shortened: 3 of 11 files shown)


src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
+"""Module energysetter of viperleed.gui.measure.
```
michele-riva: I'd suggest moving this module to the widgets package, as this is in fact a widget

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+
+    error_occurred = qtc.pyqtSignal(tuple)
+
+    def __init__(self, **kwargs):
+        """Initialize the EnergySetter widget."""
+        super().__init__(**kwargs)
+        self.energy_input = SteppingDoubleSpinBox()
+        self.set_energy = qtw.QCheckBox('Set energy')
+        self._primary_path = None
+        self._primary_controller = None
```
michele-riva: I think we can remove the "primary" bit from these names everywhere, as there is only one controller here. The notion of "primary" and "secondary" to me is only relevant for the measurement

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+        self._pending_energy = None
+        self._operation_in_progress = False
```
michele-riva: Add some comments about these internal attributes so people understand what they're used for already at init

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+    def primary_path(self, primary):
+        """Set the path to the primary controller."""
+        if primary:
+            self._primary_path = Path(primary)
+        else:
+            self._primary_path = None
+
+    def _compose(self):
+        """Set up the user interface."""
+        layout = qtw.QGridLayout(self)
```
michele-riva: Do we need a GridLayout? Perhaps a simple HorizontalLayout is enough?

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+            self._primary_path = None
+
+    def _compose(self):
+        """Set up the user interface."""
+        layout = qtw.QGridLayout(self)
+        layout.setContentsMargins(0, 0, 0, 0)
+        for widget in (self.set_energy, self.energy_input):
+            widget.setFont(AllGUIFonts().buttonFont)
+            widget.ensurePolished()
+            widget.setEnabled(False)
```
michele-riva, line 140: Do you need to setEnabled(False) here? I think this would be a weird "default state" for a freshly created widget. Also, usually it is better to disable self rather than all of its children.

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+        self.energy_input.setSuffix(' eV')
+        layout.addWidget(self.energy_input, 0, 1)
+
+        self.setLayout(layout)
+
+    def _connect(self):
+        """Connect internal signals and slots."""
+        self.set_energy.stateChanged.connect(self._on_set_energy_toggled)
+        self.energy_input.editingFinished.connect(self._on_energy_changed)
+        self.energy_input.valueChanged.connect(self._on_energy_changed)
```
michele-riva: This connection may double-fire the same energy. Perhaps not ideal? Also, I don't see a connection involving the stepped signal, so maybe you don't need the special SpinBox?

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+from viperleed.gui.measure.classes.settings import NoSettingsError
+from viperleed.gui.measure.classes.settings import ViPErLEEDSettings
+from viperleed.gui.measure.widgets.spinboxes import SteppingDoubleSpinBox
+from viperleed.gui.widgets.lib import AllGUIFonts
+
+
+class EnergySetterErrors(base.ViPErLEEDErrorEnum):
+    """Class for errors occurring in the EnergySetter widget."""
+
+    NO_PRIMARY_CONTROLLER = (2000, 'No primary controller configured. Please '
```
michele-riva: Update these error messages as per suggestions in uimeasurement

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Emits
+        -----
+        EnergySetterErrors.CONTROLLER_LOAD_FAILED
+            If making the primary controller failed.
+        EnergySetterErrors.CONTROLLER_CONNECTION_FAILED
+            If connecting the primary controller failed.
+        """
+        # If we already have a controller for this path, reuse it.
+        if self._primary_controller is not None:
```
michele-riva: There is some repetition of code in this block and further down, and there are some bits missing here from the errors further down (e.g., handling failure to reconnect). Perhaps there is a way to refactor things?

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+            ctrl_settings = self._primary_controller.settings
+            if (ctrl_settings.last_file and
+                ctrl_settings.last_file == self.primary_path):
```
michele-riva: This only checks the path, but the file contents may have actually changed. Perhaps better re-read the file (use ctrl_settings.read_again())

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+                if not self._primary_controller.connected:
+                    self._primary_controller.connect_()
+                return self._primary_controller
+            # Different controller, clean up the old one.
+            self.cleanup_primary_controller()
+
+        # Create new controller instance.
+        try:
+            primary_ctrl = self._make_primary_controller()
+        except (NoSettingsError, ValueError) as err:
```
michele-riva: I think we used `exc` for exceptions

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+    def _on_timeout(self):
+        """Handle timeout.
+
+        Emits
+        -----
+        EnergySetterErrors.SET_ENERGY_TIMEOUT
+            When a timeout happened.
+        """
+        base.emit_error(self, EnergySetterErrors.SET_ENERGY_TIMEOUT)
+        self._flush()
```
michele-riva: Probably can just call self._on_error?

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+        """Return whether the energy setter is setting energies or not."""
+        return self.set_energy.checkState() == qtc.Qt.Checked
+
+    @property
+    def primary_path(self):
+        """Return the path to the primary controller settings."""
+        return self._primary_path
+
+    @primary_path.setter
+    def primary_path(self, primary):
```
michele-riva:
```suggestion
    def primary_path(self, path):
```

michele-riva: Also, consider whether this is the right spot to update the tooltip text on the spinbox that we discussed in person

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+        except (NoSettingsError, ValueError) as err:
+            base.emit_error(self, EnergySetterErrors.CONTROLLER_LOAD_FAILED,
+                            err)
+            return None
+
+        # Connect and store the controller.
+        base.safe_connect(primary_ctrl.error_occurred, self._on_error,
+                          type=qtc.Qt.UniqueConnection)
+        base.safe_connect(primary_ctrl.serial.busy_changed,
+                          self._on_ctrl_finished, type=qtc.Qt.QueuedConnection)
```
michele-riva, line 225: Probably safer to make it also unique

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+            If the settings file cannot be read.
+        ValueError
+            If ctrl_cls_name was not found.
+        """
+        ctrl_settings = ViPErLEEDSettings()
+        ctrl_settings.read(self.primary_path)
+        ctrl_cls_name = ctrl_settings.get('controller', 'controller_class')
+        ctrl_settings.prepare_aliases(ctrl_cls_name)
+        ctrl_cls = base.class_from_name('controller', ctrl_cls_name)
+        address = ctrl_settings.get('controller', 'address')
```
michele-riva: Is this address needed? I think we get it from the settings anyway if it is not passed.

Also, could it be that the one in the settings is wrong (e.g., plug it to a different port than the one in the settings and do not click on its entry in the Devices)?

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def _connect(self):
+        """Connect internal signals and slots."""
+        self.set_energy.stateChanged.connect(self._on_set_energy_toggled)
+        self.energy_input.editingFinished.connect(self._on_energy_changed)
+        self.energy_input.valueChanged.connect(self._on_energy_changed)
+        self._timeout_timer.timeout.connect(self._on_timeout)
+
+    @qtc.pyqtSlot(int)
+    def _on_set_energy_toggled(self, state):
```
michele-riva: Alphabetic sort of methods

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+            self.cleanup_primary_controller()
+            return
+
+        # If a new energy value was queued during the operation, process it.
+        if self._pending_energy is not None:
+            energy = self._pending_energy
+            self._pending_energy = None
+            self._set_energy(energy)
+
+    def set_enabled(self, enable):
```
michele-riva: This is perhaps a bit confusing considering that all QWidgets have a setEnabled method. Perhaps you rather want to override that?

src/viperleed/gui/measure/energysetter.py
```diff
@@ -0,0 +1,328 @@
...
+        """Clean up after energy has been set."""
+        if busy:
+            return
+        self._operation_in_progress = False
+        self._timeout_timer.stop()
+
+        # If set energy is no longer toggled, we want to disconnect the ctrl.
+        if not self.set_energy.isChecked():
+            self.cleanup_primary_controller()
+            return
```
michele-riva: Is this really needed? We do keep the controller object around when we change the checked state. Or is this meant to disconnect the controller after it has set the energy to zero as a result of the state change? Then perhaps you want to call a separate `disconnect_controller` method here that does not dispose of the object as `cleanup_...` does?

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -396,6 +401,10 @@ def closeEvent(self, event):         # pylint: disable=invalid-name
 
         self._dialogs['sys_settings'].close()
         self._dialogs['firmware_upgrade'].close()
+
+        if self._ctrls['energy_setter'].set_energy.isChecked():
+            self._ctrls['energy_setter'].set_energy.setChecked(False)
+        self._ctrls['energy_setter'].cleanup_primary_controller()
```
michele-riva: I'm not sure this is the right way to do this. setChecked(False) will set the energy to zero, but the controller is deleted right after by the cleanup call. This means we do not verify that the energy is set to zero. Perhaps this should be another "retry_later" case? Let's discuss

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +511,71 @@ def _on_devices_detected(self, detected_devices):
...
+                'No controllers detected. Please ensure at least one '
+                'controller is connected and properly configured.'
+            )
+            return
+
+        # Create selection dialog
+        dialog = qtw.QDialog(self)
+        dialog.setWindowTitle('Select Primary Controller')
+        layout = qtw.QVBoxLayout(dialog)
+        label = qtw.QLabel('Select the controller to use for setting energy:')
```
michele-riva: I would be more specific here. This (as well as the usage of the word "primary") sound as if this would be the default used also during measurements.

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +511,71 @@ def _on_devices_detected(self, detected_devices):
             self._device_search_in_progress = False
             self._update_force_detect_button_state()
 
+    @qtc.pyqtSlot()
+    def _on_select_primary_controller(self):
```
michele-riva: Consider whether it would be better to make a small dedicated dialog in the energysetter module. It could simply take the controllers as init arguments.

An alternative: use the DropdownDialog from gui/dialogs

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +511,71 @@ def _on_devices_detected(self, detected_devices):
...
+
+        if dialog.exec() != qtw.QDialog.Accepted:
+            return
+        selected_index = combo.currentIndex()
+        selected_action = controller_actions[selected_index]
+        cls, info = selected_action.data()
+        address = info.more['address']
+
+        try:
+            ctrl = self._make_device(cls, info, address=address)
```
michele-riva, line 548: I see now that you are using _make_device in order for people to select which settings file to use. This is then clear enough not to bother adding another combobox with the settings files to pick from. However, I would still tell people in the final message that energies will be calibrated according to the settings in ....

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +511,71 @@ def _on_devices_detected(self, detected_devices):
...
+        if ctrl_path and ctrl_path.is_file():
+            self.system_settings.set('DEVICES', 'primary_controller',
+                                     ctrl_path.as_posix())
+            self.system_settings.update_file()
+            self._ctrls['energy_setter'].primary_path = ctrl_path.as_posix()
+            qtw.QMessageBox.information(
+                self, 'Primary Controller Set',
+                f'{selected_action.text()} is now the controller setting energies.'
+                )
+        else:
```
michele-riva, line 565: Can this even happen? This feels like the case "not ctrl" above.

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +511,71 @@ def _on_devices_detected(self, detected_devices):
...
+            qtw.QMessageBox.warning(
+                self, 'No Settings File',
+                'This controller does not have a settings file. '
+                'Please save its settings first.'
```
michele-riva: This is not so clear. Rather point them to how they can "save settings"

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -567,22 +643,16 @@ def _compose(self):
...
-        self._ctrls['energy_input'].setEnabled(False)
-
-        font = AllGUIFonts().buttonFont
-        for ctrl in ('measure', 'abort', 'set_energy', 'energy_input'):
-            self._ctrls[ctrl].setFont(font)
-            self._ctrls[ctrl].ensurePolished()
+        self._ctrls['measure'].setFont(AllGUIFonts().buttonFont)
+        self._ctrls['measure'].ensurePolished()
+        self._ctrls['abort'].setFont(AllGUIFonts().buttonFont)
+        self._ctrls['abort'].ensurePolished()
```
michele-riva: Rather keep the loop as it was in case we ever add more stuff

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -639,6 +713,15 @@ def _compose_menu(self):
         force_detect_action.triggered.connect(self.update_device_lists)
         self._ctrls['menus']['force_detect'] = force_detect_action
 
+        # Add primary controller selection action.
+        select_primary_action = devices_menu.addAction(
+            'Select Primary Controller...'
```
michele-riva: As I said already earlier, I'm not quite happy with the "Primary" here. Let's brainstorm for alternatives.

(Also, I think we capitalize only the first word in the menu entries.)

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -639,6 +713,15 @@ def _compose_menu(self):
...
+            )
+        select_primary_action.triggered.connect(
+            self._on_select_primary_controller
+            )
+        self._ctrls['menus']['select_primary'] = select_primary_action
```
michele-riva: Do we need to store this action? It doesn't look like we're ever using it again

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -1092,6 +1173,15 @@ def _on_settings_accepted(self):
...
         # one another here.
+
+        # Save the primary controller path for convenience
+        if settings_ok and self.measurement.primary_controller:
+            _path = self.measurement.primary_controller.settings.last_file
+            self.system_settings.set('DEVICES', 'primary_controller',
+                                     _path.as_posix())
+            self._ctrls['energy_setter'].primary_path = _path.as_posix()
+            self.system_settings.update_file()
+
```
michele-riva: Hmm, not sure about this. This silent modification of the system settings may undo a manual setting that users may have willingly picked. I would compare what you get here with what is stored in the settings and ask users whether they want to update (only if they differ).

This way, the "normal user" that only ever uses a single setting file will not see a difference.

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+# pylint: disable=protected-access
+# pylint: disable=redefined-outer-name
+class _FakeController:
+    """A minimal controller-like object for testing."""
+
+    def __init__(self, connected=True):
+        """Initialize fake controller."""
+        self.connected = True
+        self._connected = connected
```
michele-riva: These are a bit confusing

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+from PyQt5 import QtCore as qtc
+from PyQt5 import QtWidgets as qtw
+from pytest import fixture
+
+from viperleed.gui.measure.energysetter import EnergySetter
+
+from .mock_qt import _FakeSignal
+
+
+_ = qtw.QApplication(sys.argv)
```
michele-riva, line 26: I'd prefer if you added pytest-qt as a test dependency: 
- requirements/tests.txt
- pyproject.toml, see "tool.pytest"/"required_plugins"

then you could use pytest.mark.usefixtures('qtbot') on those that create qwidgets

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+    """Create setter with primary path."""
+    setter = EnergySetter()
+    test_path = tmp_path / 'controller.ini'
+    test_path.write_text('[controller]\n')
+    setter.primary_path = test_path
+    return setter
+
+
+@fixture
+def ctrl_setter(mocker, setter):
```
michele-riva:
```suggestion
@fixture(name='ctrl_setter')
def fixture_ctrl_setter(mocker, setter):
```

This prevents pylint messages about overriding global names. Do the same for the `setter` fixture above

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+    assert setter.energy_input is not None
+    assert setter.set_energy is not None
+    assert setter.primary_path is None
+    assert setter._primary_controller is None
+    assert setter._pending_energy is None
+    assert not setter._operation_in_progress
+
+
+def test_setting_energy_property(ctrl_setter):
```
michele-riva: Is this supposed to work also without a controller? If yes, you can use the `setter` fixture instead

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+    setter.set_enabled(True)
+
+    assert not setter.set_energy.isEnabled()
+    assert not setter.energy_input.isEnabled()
+
+
+def test_set_enabled_with_primary_path(setter):
+    """Check set_enabled enables when primary path exists."""
+    setter.set_enabled(True)
```
michele-riva: Perhaps add also 

assert not setter.set_energy.isEnabled()

before this to make sure that the method indeed changes state

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+    setter._on_set_energy_toggled(qtc.Qt.Checked)
+
+    assert not setter.set_energy.isChecked()
+    assert setter.error_occurred.emitted == 1
+
+
+def test_on_set_energy_toggled_unchecked_sets_zero(mocker, ctrl_setter):
+    """Check that unchecking sets energy to zero."""
+    ctrl_setter._set_energy = mocker.Mock()
```
michele-riva:
```suggestion
    mock_set_energy = mocker.patch.object(ctrl_setter, '_set_energy')
```

Then you can use mock_set_energy in the assertions below

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+    assert not setter.set_energy.isChecked()
+    assert setter.error_occurred.emitted == 1
+
+
+def test_on_set_energy_toggled_missing_file(tmp_path):
+    """Check error emitted when primary path file is missing."""
+    setter = EnergySetter()
+    setter.primary_path = tmp_path / 'nonexistent.ini'
+    setter.error_occurred = _FakeSignal()
```
michele-riva, line 184: use mocker.patch.object (also in other tests)

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -262,8 +262,7 @@ def __init__(self, parent=None):
         self._ctrls = {
             'measure': qtw.QPushButton('New Measurement'),
             'abort': qtw.QPushButton('Abort'),
-            'energy_input': qtw.QLineEdit(''),                                  # TODO: QDoubleSpinBox?
-            'set_energy': qtw.QPushButton('Set energy'),
+            'energy_setter': EnergySetter(),
```
michele-riva, line 266: I had in mind that we said we could use the energy spinbox to display the current energy of the measurement

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+    test_path.write_text('[controller]\n')
+    setter.primary_path = test_path
+    return setter
+
+
+@fixture
+def ctrl_setter(mocker, setter):
+    """Create setter with mocked primary controller."""
+    fake_ctrl = _FakeController()
+    setter._get_primary_controller = mocker.Mock(return_value=fake_ctrl)
```
michele-riva:
```suggestion
    mocker.patch.object(setter, '_get_primary_controller', return_value=fake_ctrl)
```

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+    ctrl_setter._on_energy_changed()
+
+    ctrl_setter._set_energy.assert_not_called()
+
+
+def test_on_energy_changed_operation_in_progress(ctrl_setter):
+    """Check energy change queued when operation in progress."""
+    ctrl_setter.set_energy.setChecked(True)
+    ctrl_setter._operation_in_progress = True
+    ctrl_setter.energy_input.setValue(75.0)
```
michele-riva: I just realized that, in principle, the spinbox allows to set energies that are beyond the capability of the controller (say, < 0 and > 1000). What happens then?

tests/gui/measure/test_energysetter.py
michele-riva: (Multiple occurrences)

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+    ctrl_setter._on_energy_changed()
+
+    # pylint: disable-next=magic-value-comparison
+    assert ctrl_setter._pending_energy == 75.0
```
michele-riva: Check also that _set_energy is not called perhaps?

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def test_on_energy_changed_sets_energy(mocker, ctrl_setter):
+    """Check energy change triggers set_energy when idle."""
+    ctrl_setter._set_energy = mocker.Mock()
+    ctrl_setter.set_energy.setChecked(True)
+    ctrl_setter._operation_in_progress = False
```
michele-riva: This should not be needed

tests/gui/measure/test_energysetter.py
michele-riva: Multiple occurrences of similar cases below

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+def test_get_primary_controller_reuses_existing(setter):
+    """Check that existing controller is reused when path matches."""
+    fake_ctrl = _FakeController()
+    fake_ctrl.settings.last_file = setter.primary_path
+    setter._primary_controller = fake_ctrl
+
+    result = setter._get_primary_controller()
+
+    assert result is fake_ctrl
+    assert fake_ctrl.connected
```
michele-riva: Is this really a good check, considering that FakeController is "connected" on init?

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+    setter = EnergySetter()
+    test_path = tmp_path / 'controller.ini'
+    test_path.write_text(
+        '[controller]\ncontroller_class = TestCtrl\naddress = COM1\n'
+        )
+    setter.primary_path = test_path
+
+    fake_settings = mocker.Mock()
+    fake_cls = mocker.Mock(return_value=_FakeController())
+    mocker.patch('viperleed.gui.measure.energysetter.ViPErLEEDSettings',
```
michele-riva: have a global `_MODULE` in this test so you can reuse it in all the patching

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+    fake_cls.assert_called()
+    assert result is not None
+
+
+def test_cleanup_primary_controller(mocker):
+    """Check controller cleanup disconnects signals."""
+    setter = EnergySetter()
+    fake_ctrl = _FakeController()
+    setter._primary_controller = fake_ctrl
+    mocker.patch('viperleed.gui.measure.energysetter.base.safe_disconnect')
```
michele-riva: Do you really need this patching?

tests/gui/measure/test_energysetter.py
```diff
@@ -0,0 +1,462 @@
...
+    setter.set_energy.setChecked.assert_called_once_with(False)
+
+
+def test_on_ctrl_finished_busy(mocker):
+    """Check ctrl finished ignores busy state."""
+    setter = EnergySetter()
+    setter._operation_in_progress = True
+    setter._timeout_timer = mocker.Mock()
+
+    setter._on_ctrl_finished(True)
```
michele-riva:
```suggestion
    setter._on_ctrl_finished(busy=True)
```

For clarity of the test. Do the same with the next one

michele-riva (changes requested):
Looks good, though I have the usual set of suggestions

src/viperleed/gui/measure/energysetter.py
FlorianDoerr: I do actually need it here. Back when I wrote this, I noticed it would trigger multiple times, so I decided to use this SpinBox. Apparently, I then proceeded to forget about changing the signal though.

src/viperleed/gui/measure/widgets/energysetter.py
FlorianDoerr: Not needed here, ctrl is necessarily a new object at this point.

src/viperleed/gui/measure/widgets/energysetter.py
FlorianDoerr: Until a controller is configured, the widget must be disabled, so yes, it should be set to False. Setting self.setEnabled is certainly possible though.

FlorianDoerr: After implementing this, I think it is a bad idea. self.energy_input should be disabled while self.set_energy is unchecked. I'll delete it again.

src/viperleed/gui/measure/widgets/energysetter.py
FlorianDoerr: Negative. As I said in the previous comments, I am not a fan of keeping the energy_input enabled while set_energy is unchecked.

src/viperleed/gui/measure/uimeasurement.py
FlorianDoerr: No, I'd fine it extremely irritating if I tried to start a measurement and the GUI keeps asking me things. I'll simply delete this.

tests/gui/measure/test_energysetter.py
FlorianDoerr: It doesn't

tests/gui/measure/test_energysetter.py
FlorianDoerr: This calls _on_set_energy_toggled under the hood, which in turn tries to _get_controller(). ctrl_setter mocks this method.

tests/gui/measure/widgets/test_energysetter.py
FlorianDoerr: As of right now we don't need it. I can add it when we make a new widget.

src/viperleed/gui/measure/controller/abc.py
```diff
@@ -695,7 +695,7 @@ def get_settings_handler(self):                                             # TO
              '</nobr> to reach a stable value after a new energy has '
              'been set. This should be calibrated for a typical step '
              'size (e.g., 0.5 eV).'),
-            ('first_settle_time', '1st energy settle time',
+            ('first_settle_time', 'E start settle time',
```
michele-riva: use the global we have in other places

src/viperleed/gui/measure/_defaults/_system_settings.ini
```diff
@@ -7,3 +7,4 @@ firmware =
 
 [DEVICES]
 live_detection =
+controller =
```
michele-riva: maybe `default_controller`?

src/viperleed/gui/measure/measurement/abc.py
```diff
@@ -237,6 +240,7 @@ def current_energy(self, new_energy):
             raise RuntimeError('Cannot set current energy '
                                'if no energy ramp is set.')
         self._energy_ramp.current_energy = new_energy
+        self.energy_changed.emit(new_energy)
```
michele-riva, line 243: Is this (and the next edit below) the right moment to emit? What if there are long intermediate steps? Then the control unit and the UI could potentially show very different values.

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +516,65 @@ def _on_devices_detected(self, detected_devices):
...
+        if not controller_actions:
+            qtw.QMessageBox.warning(
+                self, 'No Controllers Available',
+                'No controllers detected. Please ensure at least one '
+                'controller is connected and properly configured.'
+            )
+            return
+
+        # Let the user choose one of the detected controllers.
+        names = [act.text() for act in controller_actions]
```
michele-riva: Maybe rather than indexing later on, you could already build a dict here: `ctrl_name_to_data = {a.text(): a.data() for a in controller_actions}`. You should be able to pass the dict directly to the DropdownDialog, and you can later do `cls, info = ctrl_name_to_data[dropdown.selection]`

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -500,6 +516,65 @@ def _on_devices_detected(self, detected_devices):
...
+
+        _path = ctrl.settings.last_file
+        if _path and _path.is_file():
+            self.system_settings.set('DEVICES', 'controller', _path.as_posix())
+            self.system_settings.update_file()
+            self._ctrls['energy_setter'].path = _path.as_posix()
+            qtw.QMessageBox.information(
+                self, 'Controller Set', f'{selected_action.text()} is '
+                'now the controller setting energies. Energies will '
+                'be calibrated according to the controller settings.'
```
michele-riva: I would mention the path to the file in the message

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -639,6 +712,9 @@ def _compose_menu(self):
         force_detect_action.triggered.connect(self.update_device_lists)
         self._ctrls['menus']['force_detect'] = force_detect_action
 
+        select_action = devices_menu.addAction('Select controller...')
```
michele-riva: Let's discuss about the text in this action. Perhaps "Select default controller..." would be clearer?

src/viperleed/gui/measure/widgets/energysetter.py
michele-riva: Add a comment about the reason why it's set to False. Alternatively (and clearer perhaps) is to use `self._controller is not None` instead of False. Then you don't need a comment.

src/viperleed/gui/measure/widgets/energysetter.py
michele-riva: Then you also don't need the error_occurred to be unique here

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+
+    def _update_hint(self):
+        """Update the tooltip of the EnergySetter widget."""
+        if not self.path:
+            self.setToolTip('Please select a controller in the '
+                            '"Devices" menu to set an energy.')
+        else:
+            self.setToolTip('')
+
+    def cleanup_controller(self):
```
michele-riva: public before private

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+        base.safe_connect(ctrl.error_occurred, self._on_error,
+                          type=qtc.Qt.UniqueConnection)
+        base.safe_connect(ctrl.serial.busy_changed,
+                          self._on_ctrl_finished, type=qtc.Qt.QueuedConnection)
+        if not self._connect_controller(ctrl):
+            base.safe_disconnect(ctrl.error_occurred, self._on_error)
+            base.safe_disconnect(ctrl.serial.busy_changed,
+                                 self._on_ctrl_finished)
+            ctrl.disconnect_()
+            ctrl.deleteLater()
```
michele-riva: This is essentially the same as a call to cleanup_controller (except that the latter acts upon self._controller). Perhaps you could unify them?

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+
+        # Create new controller instance.
+        try:
+            ctrl = self._make_controller()
+        except (SettingsError, configparser.Error, ValueError) as exc:
+            base.emit_error(self, EnergySetterErrors.CONTROLLER_LOAD_FAILED,
+                            exc)
+            return None
+
+        # Connect and store the controller.
```
michele-riva: No storage happens here

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+        address : str
+            The address to use for the controller. An empty string
+            makes ControllerABC fall back on the value stored in
+            ``controller/address``.
+        """
+        devices = ctrl_cls().list_devices()
+        controller_name = ctrl_settings.get('controller', 'device_name',
+                                            fallback=None)
+        for device in devices:
+            detected_name = (device.more.get('name') or '')
```
michele-riva:
```suggestion
            detected_name = device.more.get('name', '')
```

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+
+        Emits
+        -----
+        EnergySetterErrors.CONTROLLER_LOAD_FAILED
+            If making the controller or loading settings failed.
+        EnergySetterErrors.CONTROLLER_CONNECTION_FAILED
+            If connecting the controller failed.
+        """
+        # If we already have a controller for this path, reuse it.
+        if self._controller is not None:
```
michele-riva, line 202: As discussed in person, you were right to not keep the controller object lingering around in the first place. We can probably simplify the code quite a bit by re-creating a fresh object. I would suggest to leave a comment in the code here, so that we're not tempted to do this again

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+        # If a new energy value was queued during the operation, process it.
+        if self._pending_energy is not None:
+            energy = self._pending_energy
+            self._pending_energy = None
+            self._set_energy(energy)
+
+    @qtc.pyqtSlot()
+    def _on_energy_changed(self):
+        """Handle energy value change."""
+        if self.set_energy.checkState() != qtc.Qt.Checked:
```
michele-riva, line 317:
```suggestion
        if not self.set_energy.isChecked():
```

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+from viperleed.gui.measure.widgets.spinboxes import SteppingDoubleSpinBox
+from viperleed.gui.widgets.lib import AllGUIFonts
+
+
+class EnergySetterErrors(base.ViPErLEEDErrorEnum):
+    """Class for errors occurring in the EnergySetter widget."""
+
+    NO_CONTROLLER = (2000, 'No controller configured. Please select one '
+                     'from the "Devices" menu using "Select Controller...".')
+    CONTROLLER_FILE_MISSING = (2001, 'The controller settings file no '
```
michele-riva: perhaps "SETTINGS_FILE_MISSING" instead?

src/viperleed/gui/measure/widgets/energysetter.py
michele-riva: Well, this implementation exactly takes care of your concern: energy_input cannot be enabled unless set_energy is enabled and it is checked.

As discussed in person, this should become the setEnabled method, and it should call super with the value that is passed in before it is altered.

src/viperleed/gui/measure/widgets/energysetter.py
```diff
@@ -0,0 +1,466 @@
...
+        energy : float
+            Energy value in eV to display.
+
+        Returns
+        -------
+        None.
+        """
+        self.energy_input.blockSignals(True)
+        self.energy_input.setValue(energy)
+        self.energy_input.blockSignals(False)
```
michele-riva, line 494: Is it really needed to block signals with setValue? You only connect to the editingFinished and stepped signals, which should not fire upon a setValue call. (I'm sure about editingFinished, less sure about stepped.)

If you need to block then it is better to do it via a context manager:

```python
with qtc.QSignalBlocker(self.energy_input):
    self.energy_input.setValue(energy)
```

tests/gui/measure/widgets/test_energysetter.py
michele-riva: I meant that I don't like the creation of a QApplication at import time. If you really don't want to add a dependency, then create the QApplication in a fixture and mark tests that need it with it. You can place the fixture in a conftest.py module (auto-loaded by pytest).

For example:

```python
@fixture
def qapp():
    """Return a valid QApplication instance."""
    return qtw.qApp() or qtw.QApplication([])
```

Let's discuss in person.

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,607 @@
...
+    mock_cleanup.assert_called_once()
+
+
+def test_get_controller_cleanup_on_different_path(mocker, setter):
+    """Check old controller is cleaned up when path changes."""
+    old_ctrl = _FakeController()
+    old_ctrl.settings.last_file = Path('/old/path.ini')
+    setter._controller = old_ctrl
+    mock_cleanup = mocker.patch.object(setter, 'cleanup_controller')
+    _ = mocker.patch.object(setter, '_make_controller',
```
michele-riva, line 174: no need to use `_ = `, you can just leave the call by itself

michele-riva: More occurrences below

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,607 @@
...
+def test_init_creates_widgets():
+    """Check that initialization creates required widgets."""
+    setter = EnergySetter()
+
+    assert setter.energy_input is not None
+    assert setter.set_energy is not None
+    assert setter.path is None
+    assert setter._controller is None
+    assert setter._pending_energy is None
+    assert not setter._operation_in_progress
```
michele-riva:
```suggestion
    assert not setter.is_busy
```

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,607 @@
...
+
+def test_on_ctrl_finished_busy(mocker):
+    """Check ctrl finished ignores busy state."""
+    setter = EnergySetter()
+    setter._operation_in_progress = True
+    mock_timeout_timer = mocker.patch.object(setter, '_timeout_timer')
+
+    setter._on_ctrl_finished(busy=True)
+
+    assert setter._operation_in_progress
```
michele-riva:
```suggestion
    assert setter.is_busy
```

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,607 @@
...
+    setter._operation_in_progress = True
+    setter._pending_energy = None
+    mock_set_energy_btn = mocker.patch.object(setter, 'set_energy')
+    mock_set_energy_btn.isChecked.return_value = True
+    mock_timeout_timer = mocker.patch.object(setter, '_timeout_timer')
+    mock_set_energy = mocker.patch.object(setter, '_set_energy')
+
+    setter._on_ctrl_finished(busy=False)
+
+    assert not setter._operation_in_progress
```
michele-riva:
```suggestion
    assert not setter.is_busy
```

michele-riva: More occurrences below

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,607 @@
...
+    ctrl_setter.energy_input.setValue(75.0)
+
+    ctrl_setter._on_energy_changed()
+
+    # pylint: disable-next=magic-value-comparison
+    assert ctrl_setter._pending_energy == 75.0
+    mock_set_energy.assert_not_called()
+
+
+def test_on_energy_changed_sets_energy(mocker, ctrl_setter):
```
michele-riva, line 447: You may want to parametrize this test for the two signals, just to check both

tests/gui/measure/widgets/test_energysetter.py
```diff
@@ -0,0 +1,607 @@
...
+
+    setter._set_energy(50.0)
+
+    assert setter._operation_in_progress
+    mock_timeout_timer.start.assert_called_once()
+
+
+def test_setting_energy_property(ctrl_setter):
+    """Check setting_energy property reflects checkbox state."""
+    ctrl_setter.set_energy.setCheckState(qtc.Qt.Unchecked)
```
michele-riva: Perhaps can use setChecked(bool) instead?

michele-riva (changes requested):
Looks good, with some small comments (+ the stuff we discussed in person)

src/viperleed/gui/measure/measurement/abc.py
FlorianDoerr: Does that really matter? In the end (within less than a second if the electronics permit it) the electronics will reach this energy. The front panel will always show wrong values anyway.

src/viperleed/gui/measure/uimeasurement.py
```diff
@@ -1307,3 +1389,4 @@ def _switch_button_enable(self, idle=False):
         self.menuBar().setEnabled(idle)
         self.statusBar().showMessage('Ready' if idle else 'Busy')
         self._update_force_detect_button_state()
+        self._ctrls['energy_setter'].setEnabled(idle)
```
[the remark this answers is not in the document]
FlorianDoerr: As of right now, the worst thing that can happen is that the controller is simply not found when the measurement settings are loaded and displayed. This means that any controller that is used to set an energy while a measurement is started cannot be used by that measurement.

PR #560, opened 2026-08-19
Add support for the py3.14 Configparser

FlorianDoerr:
The Python 3.14 Configparser changed again how the comment prefixes are stored internally. This fix extracts them from the new implementation.

### DIFF
--- a/src/viperleed/gui/measure/classes/settings.py
+++ b/src/viperleed/gui/measure/classes/settings.py
@@ -705,7 +705,7 @@ class ViPErLEEDSettings(AliasConfigParser):
         fp.write('\n')
 
     def _make_comment_regex(self):
-        """Extract and return the comment prefixes of self."""
+        """Return a regular expression for matching comments."""
         # Note that we currently pass kwargs['comment_prefixes'] = '#;'
         # on __init__, which means the comment prefixes will always be
         # # and ; at the time of writing (Py 3.15 and before).
--- a/src/viperleed/gui/measure/classes/settings.py
+++ b/src/viperleed/gui/measure/classes/settings.py
@@ -29,6 +29,7 @@ from collections.abc import Sequence
 import copy
 import os
 from pathlib import Path
+import re
 import sys
 
 from wrapt import synchronized  # thread-safety decorator
@@ -722,10 +723,17 @@ class ViPErLEEDSettings(AliasConfigParser):
             True if the line was a comment (whether it was stored
             or not).
         """
+        prefixes = None
         try:
             prefixes = self._comment_prefixes  # < PY3_13
         except AttributeError:
-            prefixes = self._prefixes.full
+            try:
+                prefixes = self._prefixes.full # = PY3_13
+            except AttributeError:
+                pass
+        if not prefixes:
+            prefixes = re.findall(r'\^\(\\?(.?)\)',
+                                  self._comments.pattern.pattern) # > PY3_13
         for prefix in prefixes:
             if line.strip().startswith(prefix):
                 if line not in self.__comments[sectname]:
--- a/src/viperleed/gui/measure/classes/settings.py
+++ b/src/viperleed/gui/measure/classes/settings.py
@@ -723,7 +723,7 @@ class ViPErLEEDSettings(AliasConfigParser):
             pattern = '|'.join(fr'^\s*({p})' for p in escaped)
             self.__prefixes = re.compile(pattern)
             return
-        self.__prefixes = self._comments.pattern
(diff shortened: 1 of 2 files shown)


src/viperleed/gui/measure/classes/settings.py
```diff
@@ -722,10 +723,17 @@ def __store_if_comment(self, line, sectname):
...
             prefixes = self._comment_prefixes  # < PY3_13
         except AttributeError:
-            prefixes = self._prefixes.full
+            try:
+                prefixes = self._prefixes.full # = PY3_13
+            except AttributeError:
+                pass
+        if not prefixes:
+            prefixes = re.findall(r'\^\(\\?(.?)\)',
+                                  self._comments.pattern.pattern) # > PY3_13
```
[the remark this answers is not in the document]
FlorianDoerr: This is wrong. The .pattern.pattern attribute returns the content of a re.compile() here, as the prefixes are sadly not stored as pure characters anymore. https://github.com/python/cpython/blob/3.14/Lib/configparser.py#L588

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -702,6 +704,27 @@ def _write_section(self, fp, section_name, section_items, delimiter):
...
+        except AttributeError:
+            try:
+                prefixes = self._prefixes.full # = PY3_13
+            except AttributeError:
+                pass
+        if prefixes:
+            # Build regex pattern to match comment lines
+            escaped = [re.escape(p) for p in prefixes]
+            pattern = '|'.join(fr'^\s*({p})' for p in escaped)
+            self.__prefixes = re.compile(pattern)
```
michele-riva: Perhaps rename to `__comment_re`? Then it is clearer what it is? Also, I would suggest to rather have this method return the compiled regex to be more explicit about the member attribute in `__init__`. Then the method could be called `_make_comment_regex`, so `__init__` would do `self.__comment_re = self._make_comment_regex()`

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -702,6 +704,27 @@ def _write_section(self, fp, section_name, section_items, delimiter):
...
+        try:
+            prefixes = self._comment_prefixes  # < PY3_13
+        except AttributeError:
+            try:
+                prefixes = self._prefixes.full # = PY3_13
+            except AttributeError:
+                pass
+        if prefixes:
+            # Build regex pattern to match comment lines
+            escaped = [re.escape(p) for p in prefixes]
```
michele-riva:
```suggestion
            escaped = (re.escape(p) for p in prefixes)
```

No need to make a list, as it's only used once.

tests/gui/measure/classes/test_settings.py
```diff
@@ -360,6 +360,46 @@ def test_all_settings_valid(self):
...
+
+    def test_comment_regex_matches_both_prefixes(self):
+        """Check that both # and ; prefixes are matched."""
+        parser = ViPErLEEDSettings()
+        # pylint: disable-next=protected-access
+        pattern = parser._ViPErLEEDSettings__prefixes
+
+        assert pattern.match('# comment') is not None
+        assert pattern.match('; comment') is not None
+        assert pattern.match('not a comment') is None
```
michele-riva:
```suggestion
        assert pattern.match('# comment')
        assert pattern.match('; comment')
        assert not pattern.match('not a comment')
```

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -702,6 +704,27 @@ def _write_section(self, fp, section_name, section_items, delimiter):
...
+            pattern = '|'.join(fr'^\s*({p})' for p in escaped)
```
michele-riva: Are the capture groups needed in the the regex?

michele-riva (changes requested):
Looks good, besides usual small suggestions

src/viperleed/gui/measure/classes/settings.py
```diff
@@ -702,6 +704,26 @@ def _write_section(self, fp, section_name, section_items, delimiter):
             fp.write(f'{key}{value}\n')
         fp.write('\n')
 
+    def _make_comment_regex(self):
+        """Extract and return the comment prefixes of self."""
```
michele-riva:
```suggestion
        """Return a regular expression for matching comments."""
```

michele-riva (changes requested):
One small change left

michele-riva (approved):

